PR #53, opened 2020-07-16
Utilize NIOConcurrencyHelpers instead of builtin Locks.swift

kirilltitov:


kirilltitov:
This change would introduce new dependency `NIO` in `LifeCycle` target, but since the package already has SwiftNIO in dependencies, I see no harm in this. I really hate duplicate code :D

ktoso:
While I agree _in principle_ that duplication is bad, we absolutely should not not go full "DRY _at any cost!_" on things. 

There's a (hidden, and painful for our future selves) cost on depending on things just because they happen to have one or two types which would be useful. Don't get me wrong, duplicating the locks file is silly and painful but it's the right thing to do for such building block libraries as Lifecycle (and metrics, logging et al). There's a few reasons pulling a dependency one does not need is bad, but let's look at the easiest to observe one, binary size:

```swift
// .target(name: "SAMPLE", dependencies: ["Lifecycle"]),

//         .target(name: "Lifecycle", dependencies: ["Logging", "Metrics", "Backtrace"]),
-> % ls -lah .build/release/SAMPLE
-rwxr-xr-x  1 ktoso  staff   447K Jul 16 21:29 .build/release/SAMPLE

//        .target(name: "Lifecycle", dependencies: ["NIO", "Logging", "Metrics", "Backtrace"]),
-> % ls -lah .build/release/SAMPLE
-rwxr-xr-x  1 ktoso  staff   3.6M Jul 16 21:30 .build/release/SAMPLE
```

That's a pretty big difference for a dependency we're not really using! 😉 There's other, harder to "point at" problems including what if a runtime really wants to use Lifecycle but for whatever reason does not want NIO (sure, most apps end up depending on it, but should ALL systems present and future?), or other resource constrained envs; for Lifecycle to be universally adopted it must absolutely pull the least dependencies possible. Another reason that's more "human factor" than the others is that once NIO is in the Lifecycle module (which was on purpose explicitly designed to _not_ depend on it) it is bound to happen that we'd accidentally depend in API or functionality on some NIO type, making it hard to remove when/if necessary. Changes in the NIOCompat package are expected, but why should a change in NIO force any change in a module that does not really use it.

So... that's a long winded (hopefully putting some light on the context in which this decision to not depend on NIO in this module was made) way of saying: no, I don't think we can or should depend on NIO in the Lifecycle module (the NIOCompat module is of course welcome and absolutely right to depend on it).

This does not change the fact that SPM will check out NIO when this library is depended on, but that's a different matter somewhat...

kirilltitov:
Well yeah, it's reasonable... But the problem is getting worse, and it's not blind DRY following — I'm seeing this Locks.swift file in at least three repositories (excluding NIO), and this class is far from perfection (compare local and NIOs). Can't we have an independent repo/package with locks then?

ktoso:
We have discussed this a few times and we are explicitly (in consultation with the core team) opting to _not_ provide a a concurrency or locks package which would likely be replaced by future work inside the swift standard library itself.

Specifically, NIO atomics and low level concurrency things are going to be replaced with the standard: https://github.com/apple/swift-se-0282-experimental

kirilltitov:
Oh, I see.

ktoso:
Thanks for the PR and discussion though! :-)

PR #55, opened 2020-07-16
Allow zero-task startup

### LINKED ISSUE #54: Crash on start if no tasks have been registered
If we use `ServiceLifecycle` without registering any tasks, a precondition fails in `ComponentLifecycle._start`.

```swift
        precondition(tasks.count > 0, "invalid number of tasks, must be > 0")
```

I think this case should be handled and should not lead to a crash. The `ServiceLifecycle` may be used in tests in which no tasks are registered, but the class, that is tested, still expects to be initialized with a `ServiceLifecycle`.

stevapple:
Fix #54

### DIFF
--- a/Sources/Lifecycle/Lifecycle.swift
+++ b/Sources/Lifecycle/Lifecycle.swift
@@ -357,7 +357,9 @@ public class ComponentLifecycle: LifecycleTask {
     // MARK: - private
 
     private func _start(on queue: DispatchQueue, tasks: [LifecycleTask], callback: @escaping (Error?) -> Void) {
-        precondition(tasks.count > 0, "invalid number of tasks, must be > 0")
+        if tasks.count == 0 {
+            log(level: .notice, "No task provided.")
+        }
         self.stateLock.withLock {
             guard case .idle = self.state else {
                 preconditionFailure("invalid state, \(self.state)")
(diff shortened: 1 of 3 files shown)


ktoso:

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -357,7 +357,9 @@ public class ComponentLifecycle: LifecycleTask {
     // MARK: - private
 
     private func _start(on queue: DispatchQueue, tasks: [LifecycleTask], callback: @escaping (Error?) -> Void) {
-        precondition(tasks.count > 0, "invalid number of tasks, must be > 0")
+        if tasks.count == 0 {
+            log(level: .notice, "No task provided.")
+        }
```
ktoso, line 362: The order of log messages will be a bit meh here with this change:

> No task provided.
> starting

and the message style is slightly off; we start with lowercase letters and no trailing `.` in messages.

Could you move this below the log("starting")?

Message perhaps `no tasks provided`?

ktoso:
Build failure because of formatting:

```

00:00:04.032 -            log(level: .notice, "No task provided.")
00:00:04.032 +            self.log(level: .notice, "No task provided.")
00:00:04.032          }
```

please run `./scripts/sanity.sh` to kick off the formatter. Seems we didn't document this step in contributing, will add.

ktoso:
@swift-server-bot add to whitelist

PR #57, opened 2020-07-17
+readme add more info about the sanity script to the contributing readme

ktoso:
I noticed we don't really explain the script in the contributing guide... WDYT @tomerd ?


ktoso:
Should we replicate this to other repos (metrics etc)? I'll do that later

PR #58, opened 2020-07-17
Update ServiceLifecycle.Configuration comments

Buratti:
It seems to me that the comments of shutdownSignal and installBacktrace were inverted, I apologize in advance if I misunderstood something.


yim-lee:

PR #63, opened 2020-07-20
Updated README.md to include logger configuration option

darrellroot:
The README.md only listed 3 of the 4 Configuration options and did not mention the logging option.  This is important because using swift-service-lifecycle in an existing app which already uses logging can impact that logging unless this configuration item is set.

Added logging option to the configuration discussion.
Placed it first, which is the same as the argument order for ServiceLifecycle.Configuration.


yim-lee:

ktoso:
Thanks! 
// assigned to milestone

PR #70, opened 2020-08-31
expose the internal lifecycle for composition

tomerd:
motivation: application frameworks thae use ServiceLifecycle may need to expose their internal lifecycle so they can be also composed into larger systems

changes:
* expose ServiceLifecycle::underlying
* make sure backtrace installer is only called once per process
* add swift-atomics dependency
* make min swift version 5.1 to match new swift-atomics dependency

### DIFF
--- a/Sources/Lifecycle/Lifecycle.swift
+++ b/Sources/Lifecycle/Lifecycle.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftServiceLifecycle open source project
 //
-// Copyright (c) 2019-2020 Apple Inc. and the SwiftServiceLifecycle project authors
+// Copyright (c) 2019-2021 Apple Inc. and the SwiftServiceLifecycle project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
@@ -20,6 +20,7 @@ import Glibc
 import Backtrace
 import Dispatch
 import Logging
+import Metrics
 
 // MARK: - LifecycleTask
 
@@ -99,13 +100,15 @@ public struct LifecycleHandler {
 /// `ServiceLifecycle` provides a basic mechanism to cleanly startup and shutdown the application, freeing resources in order before exiting.
 ///  By default, also install shutdown hooks based on `Signal` and backtraces.
 public struct ServiceLifecycle {
+    private static let backtracesInstalled = AtomicBoolean(false)
+
     private let configuration: Configuration
 
     /// The underlying `ComponentLifecycle` instance
     ///
     /// Designed for composition purposes, mainly for frameworks that need to offer both top-level start/stop functionality and composition into larger systems.
     /// In other words, should not be used outside the context of building an Application framework.
-    private let underlying: ComponentLifecycle
+    public let underlying: ComponentLifecycle
 
     /// Creates a `ServiceLifecycle` instance.
     ///
@@ -114,11 +117,8 @@ public struct ServiceLifecycle {
     public init(configuration: Configuration = .init()) {
         self.configuration = configuration
         self.underlying = ComponentLifecycle(label: self.configuration.label, logger: self.configuration.logger)
-        // setup backtrace trap as soon as possible
-        if configuration.installBacktrace {
-            self.log("installing backtrace")
-            Backtrace.install()
-        }
+        // setup backtraces as soon as possible, so if we crash during setup we get a backtrace
+        self.installBacktrace()
     }
 
     /// Starts the provided `LifecycleTask` array.
@@ -127,6 +127,9 @@ public struct ServiceLifecycle {
     /// - parameters:
     ///    - callback: The handler which is called after the start operation completes. The parameter will be `nil` on success and contain the `Error` otherwise.
     public func start(_ callback: @escaping (Error?) -> Void) {
+        guard self.underlying.idle else {
+            preconditionFailure("already started")
+        }
         self.setupShutdownHook()
         self.underlying.start(on: self.configuration.callbackQueue, callback)
     }
@@ -134,6 +137,9 @@ public struct ServiceLifecycle {
     /// Starts the provided `LifecycleTask` array and waits (blocking) until a shutdown `Signal` is captured or `shutdown` is called on another thread.
     /// Startup is performed in the order of items provided.
     public func startAndWait() throws {
+        guard self.underlying.idle else {
+            preconditionFailure("already started")
+        }
         self.setupShutdownHook()
         try self.underlying.startAndWait(on: self.configuration.callbackQueue)
     }
@@ -150,6 +156,13 @@ public struct ServiceLifecycle {
     /// Waits (blocking) until shutdown `Signal` is captured or `shutdown` is invoked on another thread.
     public func wait() {
         self.underlying.wait()
+    }
+
+    private func installBacktrace() {
+        if self.configuration.installBacktrace, ServiceLifecycle.backtracesInstalled.compareAndSwap(expected: false, desired: true) {
+            self.log("installing backtrace")
+            Backtrace.install()
+        }
     }
 
     private func setupShutdownHook() {
@@ -351,6 +364,16 @@ public class ComponentLifecycle: LifecycleTask {
     /// Waits (blocking) until `shutdown` is invoked on another thread.
     public func wait() {
         self.shutdownGroup.wait()
+    }
+
+    // MARK: - internal
+
+    internal var idle: Bool {
+        if case .idle = self.state {
+            return true
+        } else {
+            return false
+        }
     }
 
     // MARK: - private
--- a/Tests/LifecycleTests/ServiceLifecycleTests.swift
+++ b/Tests/LifecycleTests/ServiceLifecycleTests.swift
@@ -14,6 +14,7 @@
 
 @testable import Lifecycle
 import LifecycleNIOCompat
+import Logging
 import XCTest
 
 final class ServiceLifecycleTests: XCTestCase {
@@ -225,5 +226,11 @@ final class ServiceLifecycleTests: XCTestCase {
         XCTAssertEqual("\(ServiceLifecycle.Signal.TERM)", "Signal(TERM, rawValue: \(ServiceLifecycle.Signal.TERM.rawValue))")
         XCTAssertEqual("\(ServiceLifecycle.Signal.INT)", "Signal(INT, rawValue: \(ServiceLifecycle.Signal.INT.rawValue))")
         XCTAssertEqual("\(ServiceLifecycle.Signal.ALRM)", "Signal(ALRM, rawValue: \(ServiceLifecycle.Signal.ALRM.rawValue))")
+    }
+
+    func testBacktracesInstalledOnce() {
+        let config = ServiceLifecycle.Configuration(installBacktrace: true)
+        _ = ServiceLifecycle(configuration: config)
+        _ = ServiceLifecycle(configuration: config)
     }
 }
--- a/Package.swift
+++ b/Package.swift
@@ -14,9 +14,24 @@ let package = Package(
         .package(url: "https://github.com/swift-server/swift-backtrace.git", from: "1.1.1"),
         .package(url: "https://github.com/apple/swift-nio.git", from: "2.0.0"), // used in tests
     ],
-    targets: [
-        .target(name: "Lifecycle", dependencies: ["Logging", "Metrics", "Backtrace"]),
-        .target(name: "LifecycleNIOCompat", dependencies: ["Lifecycle", "NIO"]),
-        .testTarget(name: "LifecycleTests", dependencies: ["Lifecycle", "LifecycleNIOCompat"]),
-    ]
+    targets: []
 )
+
+#if compiler(>=5.2)
+package.dependencies += [
+    .package(url: "https://github.com/apple/swift-atomics.git", .exact("0.0.3")), // exact since < 1.0
+]
+package.targets += [
+    .target(name: "Lifecycle", dependencies: ["Logging", "Metrics", "Backtrace", "Atomics"]),
+]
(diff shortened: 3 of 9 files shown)


Sources/Lifecycle/Lifecycle.swift
```diff
@@ -83,61 +83,66 @@ public struct LifecycleHandler {
 
 public struct ServiceLifecycle {
     private let configuration: Configuration
-    private let lifecycle: ComponentLifecycle
 
-    /// Creates a `ComponentLifecycle` instance.
+    /// The internal `ComponentLifecycle` instance
+    ///
+    /// Designed for composition purposes, mainly for frameworks that need to offer both top-level strart/stop functionality and composition into larger systems.
```
yim-lee, line 89:
```suggestion
    /// Designed for composition purposes, mainly for frameworks that need to offer both top-level start/stop functionality and composition into larger systems.
```

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -83,61 +83,66 @@ public struct LifecycleHandler {
...
+    /// In other words, should not be in used outside the context of building an Application framework.
```
yim-lee, line 90:
```suggestion
    /// In other words, should not be used outside the context of building an Application framework.
```

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -186,45 +191,47 @@ extension ServiceLifecycle {
...
+        self.guts.register(tasks)
     }
 
-    /// Adds a `Task` to a `Tasks` collection.
+    /// Adds a `LifecycleTask` to a `LifecycleTasks` collection.
     ///
     /// - parameters:
     ///    - label: label of the item, useful for debugging.
     ///    - start: `Handler` to perform the startup.
     ///    - shutdown: `Handler` to perform the shutdown.
```
yim-lee, line 215:
```suggestion
    ///    - shutdown: `LifecycleHandler` to perform the shutdown.
```

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -186,45 +191,47 @@ extension ServiceLifecycle {
...
-        self.lifecycle.register(tasks)
+        self.guts.register(tasks)
     }
 
-    /// Adds a `Task` to a `Tasks` collection.
+    /// Adds a `LifecycleTask` to a `LifecycleTasks` collection.
     ///
     /// - parameters:
     ///    - label: label of the item, useful for debugging.
     ///    - start: `Handler` to perform the startup.
```
yim-lee, line 214:
```suggestion
    ///    - start: `LifecycleHandler` to perform the startup.
```

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -186,45 +191,47 @@ extension ServiceLifecycle {
...
-        self.lifecycle.register(label: label, start: start, shutdown: shutdown)
+        self.guts.register(label: label, start: start, shutdown: shutdown)
     }
 
-    /// Adds a `Task` to a `Tasks` collection.
+    /// Adds a `LifecycleTask` to a `LifecycleTasks` collection.
     ///
     /// - parameters:
     ///    - label: label of the item, useful for debugging.
     ///    - handler: `Handler` to perform the shutdown.
```
yim-lee, line 224:
```suggestion
    ///    - handler: `LifecycleHandler` to perform the shutdown.
```

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -482,18 +491,18 @@ public extension ComponentLifecycle {
         }
     }
 
-    /// Adds a `Task` to a `Tasks` collection.
+    /// Adds a `LifecycleTask` to a `LifecycleTasks` collection.
     ///
     /// - parameters:
-    ///    - tasks: one or more `Tasks`.
+    ///    - tasks: one or more `LifecycleTasks`.
```
yim-lee, line 497:
```suggestion
    ///    - tasks: one or more `LifecycleTask`.
```

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -482,18 +491,18 @@ public extension ComponentLifecycle {
...
     func register(_ tasks: LifecycleTask ...) {
         self.register(tasks)
     }
 
-    /// Adds a `Task` to a `Tasks` collection.
+    /// Adds a `LifecycleTask` to a `LifecycleTasks` collection.
     ///
     /// - parameters:
-    ///    - tasks: array of `Tasks`.
+    ///    - tasks: array of `LifecycleTasks`.
```
yim-lee, line 505:
```suggestion
    ///    - tasks: array of `LifecycleTask`.
```

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -83,61 +83,66 @@ public struct LifecycleHandler {
...
 public struct ServiceLifecycle {
     private let configuration: Configuration
-    private let lifecycle: ComponentLifecycle
 
-    /// Creates a `ComponentLifecycle` instance.
+    /// The internal `ComponentLifecycle` instance
+    ///
+    /// Designed for composition purposes, mainly for frameworks that need to offer both top-level strart/stop functionality and composition into larger systems.
+    /// In other words, should not be in used outside the context of building an Application framework.
+    public let guts: ComponentLifecycle
```
yim-lee, line 91: I suggest `_lifecycle` instead of `guts` but not a strong opinion.

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -186,45 +191,47 @@ extension ServiceLifecycle {
 }
 
 public extension ServiceLifecycle {
-    /// Adds a `Task` to a `Tasks` collection.
+    /// Adds a `LifecycleTask` to a `LifecycleTasks` collection.
     ///
     /// - parameters:
-    ///    - tasks: one or more `Tasks`.
+    ///    - tasks: one or more `LifecycleTasks`.
```
yim-lee, line 197:
```suggestion
    ///    - tasks: one or more `LifecycleTask`.
```

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -186,45 +191,47 @@ extension ServiceLifecycle {
...
-        self.lifecycle.register(tasks)
+        self.guts.register(tasks)
     }
 
-    /// Adds a `Task` to a `Tasks` collection.
+    /// Adds a `LifecycleTask` to a `LifecycleTasks` collection.
     ///
     /// - parameters:
-    ///    - tasks: array of `Tasks`.
+    ///    - tasks: array of `LifecycleTasks`.
```
yim-lee, line 205:
```suggestion
    ///    - tasks: array of `LifecycleTask`.
```

Sources/Lifecycle/Lifecycle.swift
tomerd: this is totally open to opinions

ktoso: Yeah... guts is a bit weird :) 
Yim's `_lifecycle` suggestion is good, or `underlying` which I use often as well?

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -83,61 +83,66 @@ public struct LifecycleHandler {
 
 public struct ServiceLifecycle {
     private let configuration: Configuration
-    private let lifecycle: ComponentLifecycle
 
-    /// Creates a `ComponentLifecycle` instance.
+    /// The internal `ComponentLifecycle` instance
```
ktoso, line 87:
```suggestion
    /// The underlying `ComponentLifecycle` instance
```

to avoid `internal` word which can be confusing?

Sources/Lifecycle/Lifecycle.swift
ktoso: `_` I guess would imply that it's not "stable" so perhaps `underlying` or `componentLifecycle` after all?
I suppose this is effectively stable API after all right?

ktoso (approved):
Minor naming nitpick but LGTM otherwise

Sources/Lifecycle/Lifecycle.swift
avolokhov: If I read the code correctly, having it public will mean we allow multiple `ServiceLifecycle` instances to be initialised in the application simultaneously. Would it lead to multiple backtraces signal handlers being set up? I see it's a default behaviour for `ServiceLifecycle`.

tomerd: @ktoso @yim-lee `underlying` works for me

tomerd: @avolokhov the goal here is not to allow multiple ServiceLifecycle instances to be initialised in the application simultaneously, the goal is to enable composition, see [exchange with @tanner0101](https://forums.swift.org/t/sswg-0015-swift-service-lifecycle/39157/6?u=tomerd) for background

avolokhov: Thanks for the context! Sorry if I'm about to talk nonsense, I'm trying to better understand the use case.
According to the sample in that thread
```
let lifecycle = ServiceLifecycle()

let app = try Application(.detect())
// app.lifecycle is ComponentLifecycle
lifecycle.register(app.lifecycle)

// other stuff

try lifecycle.startAndWait()
```
`app.lifecycle` here is a `ComponentLifecycle`, which is also `guts` (or `underlying`) for some `ServiceLifecycle` defined inside `Application`? In this case, `Application` doesn't know in advance whether it'll be used as a part of a bigger `Lifecycle` or as a top level app, so it'll have to install backtraces when creating its own `ServiceLifecycle`.
Then if it's used as a part of a bigger `Lifecycle` this bigger `Lifecycle` will have to install backtraces on its own. This will lead to multiple backtraces handlers set up in a single application. If I read it correctly, to avoid it we'll have to install backtraces handler on `start`, rather than on `init`?

I tried to follow the case @tanner0101 described in that thread https://forums.swift.org/t/sswg-0015-swift-service-lifecycle/39157/7
And got the following sample code:

```
    struct Application {
        let appLifecycle: ServiceLifecycle
        init() {
            appLifecycle = ServiceLifecycle()
        }

        var lifecycle: ComponentLifecycle {
            appLifecycle.underlying
        }

        func start() {
            appLifecycle.start {_ in }
        }

        func shutdown() {
            appLifecycle.shutdown()
        }
    }



    func testPartOfBiggerApp() throws {
        let lifecycle = ServiceLifecycle()

        let app = try Application()
        // app.lifecycle is ComponentLifecycle
        lifecycle.register(app.lifecycle)

        // other stuff

        try lifecycle.startAndWait()
    }
```
In this sample code two backtrace handlers are installed: one in `Application. appLifecycle.init()`, the other in `testPartOfBiggerApp` in `lifecycle.init()`. In general, my understanding is that a call to `lifecycle.underlying` implies that this `lifecycle` is created, and therefore backtrace handler is installed. When we use it as a `LifecycleTask` for some other `ServiceLifecycle`, we assume that the other `ServiceLifecycle` exists and therefore installed backtrace handler again.

tomerd: @avolokhov great catch, this is in fact a problem with the proposed solution 

I think I will modify this PR to address the other small improvements it includes but revert `underlying` to be private until we have a better solution

cc @tanner0101

tomerd: > If I read it correctly, to avoid it we'll have to install backtraces handler on start, rather than on init?

this could be one solution yes. the reason we had it in `init` originally was to install them as soon as possible as the application may crash before calling start (from real-world experience)

tomerd: another solution could be for application frameworks like vapor to take in an optional `Lifecycle` in their constructor

tomerd: @avolokhov finally got back to this PR... updated to moving the backtrace installer to `start` as you suggested

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -152,6 +147,15 @@ public struct ServiceLifecycle {
         self.underlying.wait()
     }
 
+    private func setupForStart() {
```
tomerd, line 150: we could consider an assertion here based on a static to make sure theses never called twice since that may lead to undefined behavior. wdyt?

tomerd:
@yim-lee @ktoso @avolokhov

tomerd:

tomerd:

tomerd:

Tests/LifecycleTests/ServiceLifecycleTests.swift
```diff
@@ -226,4 +227,10 @@ final class ServiceLifecycleTests: XCTestCase {
         XCTAssertEqual("\(ServiceLifecycle.Signal.INT)", "Signal(INT, rawValue: \(ServiceLifecycle.Signal.INT.rawValue))")
         XCTAssertEqual("\(ServiceLifecycle.Signal.ALRM)", "Signal(ALRM, rawValue: \(ServiceLifecycle.Signal.ALRM.rawValue))")
     }
+
+    func testBacktracesInstalledOnce() {
+        let config = ServiceLifecycle.Configuration(installBacktrace: true)
+        _ = ServiceLifecycle(configuration: config)
+        _ = ServiceLifecycle(configuration: config)
```
yim-lee, line 234: duplicate?

tomerd: deliberate... making sure second instantiation does not trigger another setup of backtraces

yim-lee: i see. maybe add a code comment.

yim-lee (approved):

PR #73, opened 2020-09-01
API docs cleanup and other minor fixes

tomerd:
motivation: better API docs and usability

changes:
* allow customization of ServiceLifecycle label
* adjust APi docs
* fix tests on macOS (availability)


tomerd:
reviewed as part of #70

PR #75, opened 2020-09-18
abstract registration function into a protocol

tomerd:
motivation: allow libraries that want to expose the lifecycle so that downstream code could register additional items to do so regardless of the lifecycle backend

changes: abstract registration function into a protocol and conform ComponentLifecycle and ServiceLifecycle to it

### DIFF
--- a/Sources/Lifecycle/Lifecycle.swift
+++ b/Sources/Lifecycle/Lifecycle.swift
@@ -194,40 +194,9 @@ extension ServiceLifecycle {
     }
 }
 
-public extension ServiceLifecycle {
-    /// Adds a `LifecycleTask` to a `LifecycleTasks` collection.
-    ///
-    /// - parameters:
-    ///    - tasks: one or more `LifecycleTask`.
-    func register(_ tasks: LifecycleTask ...) {
+extension ServiceLifecycle: LifecycleRegistrar {
+    public func register(_ tasks: [LifecycleTask]) {
         self.underlying.register(tasks)
-    }
-
-    /// Adds a `LifecycleTask` to a `LifecycleTasks` collection.
-    ///
-    /// - parameters:
-    ///    - tasks: array of `LifecycleTask`.
-    func register(_ tasks: [LifecycleTask]) {
-        self.underlying.register(tasks)
-    }
-
-    /// Adds a `LifecycleTask` to a `LifecycleTasks` collection.
-    ///
-    /// - parameters:
-    ///    - label: label of the item, useful for debugging.
-    ///    - start: `LifecycleHandler` to perform the startup.
-    ///    - shutdown: `LifecycleHandler` to perform the shutdown.
-    func register(label: String, start: LifecycleHandler, shutdown: LifecycleHandler) {
-        self.underlying.register(label: label, start: start, shutdown: shutdown)
-    }
-
-    /// Adds a `LifecycleTask` to a `LifecycleTasks` collection.
-    ///
-    /// - parameters:
-    ///    - label: label of the item, useful for debugging.
-    ///    - handler: `LifecycleHandler` to perform the shutdown.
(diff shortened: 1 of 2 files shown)


Sources/Lifecycle/Lifecycle.swift
```diff
@@ -480,42 +449,34 @@ public class ComponentLifecycle: LifecycleTask {
...
-
-        func shutdown(_ callback: @escaping (Error?) -> Void) {
-            self.shutdown.run(callback)
+        self.tasksLock.withLock {
+            self.tasks.append(contentsOf: tasks)
         }
     }
+}
 
+public protocol LifecycleRegistrar {
```
tomerd, line 465: not in love with this name, better ideas?

tomerd:
@fabianfett as discussed for the lambda use case

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -524,16 +485,30 @@ public extension ComponentLifecycle {
...
     /// - parameters:
     ///    - label: label of the item, useful for debugging.
     ///    - handler: `Handler` to perform the shutdown.
-    func registerShutdown(label: String, _ handler: LifecycleHandler) {
+    public func registerShutdown(label: String, _ handler: LifecycleHandler) {
         self.register(label: label, start: .none, shutdown: handler)
     }
 }
+
+internal struct LifecycleTaskImpl: LifecycleTask {
```
yim-lee, line 502: Is it more common to name these `LifecycleTask: LifecycleTaskProtocol`?  `*Impl` is kind of Java-y.

tomerd: `LifecycleTask` is the public "currency type", while `LifecycleTaskImpl` is internal. While there are a few examples where Swift stdlib uses the `XxxProtocol` convention (e.g. `StringProtocol`) its actually not all that common to use it for public "currency type". That said I too not in love with `LifecycleTaskImpl` but since its internal I let it slide.

fabianfett:
@tomerd Since we originally intended this change for Lambda, I think we should first find a solution to this: https://github.com/swift-server/swift-aws-lambda-runtime/pull/141#issuecomment-698555030

Or do you want this change no matter what Lambda does?

tomerd:
> Or do you want this change no matter what Lambda does? 

I think this is a good change in any case so I would merge this

Sources/Lifecycle/Lifecycle.swift
ktoso: It's not too bad if it's internal; true that making the actual public name good is more important 👍

fabianfett (approved):

tomerd:
@yim-lee @fabianfett @ktoso 

renamed `LifecycleRegistrar`  -> `LifecycleTasksConainer`  and `LifecycleTaskImpl` -> `_ LifecycleTask`

wdyt?

yim-lee (approved):
Typo in type name, otherwise 👍

ktoso:
sounds good 👍

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -194,41 +196,10 @@ extension ServiceLifecycle {
...
-    func register(_ tasks: LifecycleTask ...) {
-        self.underlying.register(tasks)
-    }
-
-    /// Adds a `LifecycleTask` to a `LifecycleTasks` collection.
-    ///
-    /// - parameters:
-    ///    - tasks: array of `LifecycleTask`.
-    func register(_ tasks: [LifecycleTask]) {
+extension ServiceLifecycle: LifecycleTasksConainer {
```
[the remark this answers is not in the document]
ktoso: LifecycleTaskContainer sounds good :)

ktoso (approved):

Issue #21, opened 2020-04-03
should be allow specifying a DispatchQueue per registered item?


weissi:
Often, sub-systems won't be thread-safe so we rely on start & stop being called from the same thread. That is achievable with `Configuration` however only per `Lifecycle` and not once per item. What if two separate items rely on being each started/stopped on their own queue?

FranzBusch:
Since we just changed the whole package to use a new Structured Concurrency approach, I would like to close this issue. @weissi Is that fine with you?

PR #77, opened 2020-10-09
add semantics for stopping tasks that have no start handler

tomerd:
motivation: address shutdown for tasks which allocate resources on construction. before this change, only tasks that have been started are shutdown, this creates an issue when the resources are allocated on construction since the item may not get started at all (for example due to error in earlier task) and thus will never be shutdown and the resources will leak.

changes: add semantics to request shutdown even if not started, and assume this se semantics when registering with `start: .none` or with `registerShutdown`

### DIFF
--- a/Sources/Lifecycle/Lifecycle.swift
+++ b/Sources/Lifecycle/Lifecycle.swift
@@ -256,11 +256,8 @@ public class ComponentLifecycle: LifecycleTask {
     private let logger: Logger
     internal let shutdownGroup = DispatchGroup()
 
-    private var state = State.idle
+    private var state = State.idle([])
     private let stateLock = Lock()
-
-    private var tasks = [LifecycleTask]()
-    private let tasksLock = Lock()
 
     /// Creates a `ComponentLifecycle` instance.
     ///
@@ -289,7 +286,9 @@ public class ComponentLifecycle: LifecycleTask {
     ///    - on: `DispatchQueue` to run the handlers callback  on
     ///    - callback: The handler which is called after the start operation completes. The parameter will be `nil` on success and contain the `Error` otherwise.
     public func start(on queue: DispatchQueue, _ callback: @escaping (Error?) -> Void) {
-        let tasks = self.tasksLock.withLock { self.tasks }
+        guard case .idle(let tasks) = (self.stateLock.withLock { self.state }) else {
+            preconditionFailure("invalid state, \(self.state)")
+        }
         self._start(on: queue, tasks: tasks, callback: callback)
     }
 
@@ -325,11 +324,17 @@ public class ComponentLifecycle: LifecycleTask {
 
         self.stateLock.lock()
         switch self.state {
-        case .idle:
+        case .idle(let tasks) where tasks.isEmpty:
             self.state = .shutdown(nil)
             self.stateLock.unlock()
             defer { self.shutdownGroup.leave() }
             callback(nil)
+        case .idle(let tasks):
+            self.stateLock.unlock()
+            let queue = DispatchQueue.global() // is this the best we can do?
+            setupShutdownListener(queue)
+            let stoppable = tasks.filter { $0.shutdownIfNotStarted }
+            self._shutdown(on: queue, tasks: stoppable, callback: self.shutdownGroup.leave)
         case .shutdown:
             self.stateLock.unlock()
             self.log(level: .warning, "already shutdown")
@@ -342,7 +347,6 @@ public class ComponentLifecycle: LifecycleTask {
             self.stateLock.unlock()
             setupShutdownListener(queue)
         case .started(let queue, let tasks):
-            self.state = .shuttingDown(queue)
             self.stateLock.unlock()
             setupShutdownListener(queue)
             self._shutdown(on: queue, tasks: tasks, callback: self.shutdownGroup.leave)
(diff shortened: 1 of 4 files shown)


Sources/Lifecycle/Lifecycle.swift
```diff
@@ -26,22 +26,32 @@ import Logging
...
     private let body: (@escaping (Error?) -> Void) -> Void
+    internal let noop: Bool
 
     /// Initialize a `LifecycleHandler` based on a completion handler.
     ///
     /// - parameters:
     ///    - callback: the underlying completion handler
-    public init(_ callback: @escaping (@escaping (Error?) -> Void) -> Void) {
+    ///    - noop: the underlying completion handler is a no-op
+    public init(_ callback: @escaping (@escaping (Error?) -> Void) -> Void, noop: Bool = false) {
```
yim-lee, line 52: Not sure I understand what `noop` does. It seems a bit unsafe to let user set this?

tomerd: good point may this private

yim-lee (approved):

Tests/LifecycleTests/ComponentLifecycleTests.swift
```diff
@@ -839,4 +840,148 @@ final class ComponentLifecycleTests: XCTestCase {
...
+        lifecycle.register(label: item3.label, start: .sync(item3.start), shutdown: .sync(item3.shutdown))
+
+        let item4 = DestructionSensitive(label: "4", sempahore: sempahore)
+        lifecycle.registerShutdown(label: item4.label, .sync(item4.shutdown))
+
+        let item5 = DestructionSensitive(label: "5", sempahore: sempahore)
+        lifecycle.register(label: item5.label, start: .none, shutdown: .sync(item5.shutdown))
+
+        let item6 = DestructionSensitive(label: "6", sempahore: sempahore)
+        lifecycle.register(label: item6.label, shutdownIfNotStarted: true, start: .sync(item3.start), shutdown: .sync(item6.shutdown))
```
ktoso, line 977: Mixed feelings here tbh.

I would love to rather revisit the design I proposed a while ago which allows the shutdown to get values from the start. This way, if there was no start called there is no start value.

I.e. 

```swift
// start, semantically equivalent to:  
//     () -> SomeValue / EventLoopFuture<SomeValue>
// shutdown, semantically equivalent to: 
//    (TaskStartResult<SomeValue, Error>) -> Void / EventLoopFuture<Void>
lifecycle.register(
    label: item6.label, 
    start: .sync(DestructionSensitive()), 
    shutdown: .sync { $0.success?.shutdown() }
)
```

etc... Something worth exploring? Should I prototype it or would you have a look @tomerd ?
I remain very worried about those "allocate object outside" and then mutate start/stop the thing from lifecycle callbacks without a communication channel between then somehow... (esp if threading was involved which we might get to because: https://github.com/swift-server/swift-service-lifecycle/issues/21 )

tomerd: absolutely - you are more than welcome to explore and prototype and alternative better design. note that in this case the `register` function will also need to return the instance that is created in `start` since it is needed "outside" as well.

ktoso: Ah yeah I see could be tricky... don't treat this comment as blocker -- I'll try to poke around if I could make it happen

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -311,11 +324,17 @@ public class ComponentLifecycle: LifecycleTask {
...
         switch self.state {
-        case .idle:
+        case .idle(let tasks) where tasks.isEmpty:
             self.state = .shutdown(nil)
             self.stateLock.unlock()
             defer { self.shutdownGroup.leave() }
             callback(nil)
+        case .idle(let tasks):
+            self.stateLock.unlock()
+            let queue = DispatchQueue.global() // is this the best we can do?
```
fabianfett, line 334: As of now the queue that is used in the other states is supplied during `start()`. Maybe we could supply this queue in the initializer and set it as a `let`? This way we could access the same queue for stop in any situation?

fabianfett (commented):
I think this really good change, that will make live in many use cases much simpler. Left some comments here and there for discussion. 

Before we merge we should definitely make the distinction between use cases (and in what "modes" ServiceLifecycle can be used) more visible in documentation.

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -26,22 +26,36 @@ import Logging
...
+}
+
 // MARK: - LifecycleHandler
 
 /// Supported startup and shutdown method styles
 public struct LifecycleHandler {
     private let body: (@escaping (Error?) -> Void) -> Void
+    internal let noop: Bool
+
+    private init(_ callback: @escaping (@escaping (Error?) -> Void) -> Void, noop: Bool) {
```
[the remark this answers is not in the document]
tomerd: good idea!

Sources/Lifecycle/Lifecycle.swift
tomerd: remove this shutdown semantics

tomerd:
@fabianfett simplified based on discussion, ptal

ktoso:
Admitting I did not have time to loop back here... Feel free to merge if you want to @tomerd

Issue #11, opened 2020-03-31
think about associated data when started


weissi:
Manual state tracking has the advantage that you're able to associate data with a state. For example

```swift
enum LifecycleState {
    case stopped
    case started(Channel)
    case stopping(Channel)
}
```

migrating this to `Lifecycle` currently would require to add an optional

```
var serverChannel: Channel?
```

and then the code would start to contain `self.serverChannel!`. The other option is to duplicate the state tracking in `Lifecycle` as well as the application but that's not nice either because the states can get out of sync.

rauhul:
As a new user to Lifecycle, I found the need to erase my EventLoopFuture<Channel> to EventLoopFuture<Void> a bit cumbersome/unintuitive. I had to store the result of start in a sideband promise so I could properly shutdown the Channel.

ex:
```swift
final class MetricsHTTPServer {
    let lifecycle = ComponentLifecycle(label: "foo")

    private(set) var channelPromise: EventLoopPromise<Channel>
    let eventLoopGroup: EventLoopGroup

    init(eventLoopGroup: EventLoopGroup) {
        self.eventLoopGroup = eventLoopGroup
        self.channelPromise = eventLoopGroup.next().makePromise(of: Channel.self)
        lifecycle.register(
            label: "Channel",
            start: .eventLoopFuture(self.start),
            shutdown: .eventLoopFuture(self.shutdown))
    }

    func start() -> EventLoopFuture<Void> {
        let bootstrap = ServerBootstrap(group: eventLoopGroup)
        // ...
        let future = bootstrap.bind(host: ..., port: ...)
        future.cascade(to: channelPromise)
        return future.map { _ in }
    }

    func shutdown() -> EventLoopFuture<Void> {
        return channelPromise.futureResult.flatMap { $0.close() }
    }
}
```

As a user, I would've preferred to defer the storage of the Channel to the ComponentLifecycle and have it passed back to me as part of the shutdown callback.

weissi:
I don't think this is solved. What I was looking for is a way to (from the outside) access what the `start` method constructed.

weissi:
For example, what I seem to always need is something like this:

```swift
        let service = ServiceLifecycle(configuration: .init(label: "proxy",
                                                            logger: logger,
                                                            callbackQueue: DispatchQueue.global(),
                                                            shutdownSignal: [.INT],
                                                            installBacktrace: true))
        var downstreamClient: Client?
        var otherComponent: Other?
[...]

PR #84, opened 2021-03-11
add support for stateful handler

tomerd:
motivation: allow lifecycle handlers to have the library manage the state for them so they do not need to do that manually

changes:
* introduce LifecycleStartHandler and LifecycleShutdownHandler which can handle state on behalf of the lifecycle item
* add registerStateful function to regsiter stateful handlers
* add tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,6 +214,32 @@ In more complex cases, when `Signal`-trapping-based shutdown is not appropriate,
 
 `shutdown` is an asynchronous operation. Errors will be logged and bubbled up to the provided completion handler.
 
+### Stateful handlers
+
+In some cases it is useful to have the Start handlers return a state that can be passed on to the Shutdown handlers for shutdown.
+For example, when establishing some sort of a connection that needs to be closed at shutdown.
+
+```swift
+struct Foo {
+  func start() throws -> Connection {
+      return ...
+  }
+
+  func shutdown(state: Connection) throws {
+      ...
+  }
+}
+```
+
+```swift
+let foo = ...
+lifecycle.registerStateful(
+    label: "foo",
+    start: .sync(foo.start),
+    shutdown: .sync(foo.shutdown)
+)
+```
+
 ### Complex Systems and Nesting of Subsystems
 
 In larger Applications (Services) `ComponentLifecycle` can be used to manage the lifecycle of subsystems, such that `ServiceLifecycle` can start and shutdown `ComponentLifecycle`s.
--- a/Sources/Lifecycle/Lifecycle.swift
+++ b/Sources/Lifecycle/Lifecycle.swift
@@ -41,25 +41,25 @@ extension LifecycleTask {
 
 /// Supported startup and shutdown method styles
 public struct LifecycleHandler {
+    @available(*, deprecated)
     public typealias Callback = (@escaping (Error?) -> Void) -> Void
 
-    private let body: Callback?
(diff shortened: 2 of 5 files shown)


tomerd:
one year later, but this addresses #11 if I understand the need correctly

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -82,15 +84,101 @@ public struct LifecycleHandler {
...
-        return self.body == nil
+        return self.underlying == nil
+    }
+}
+
+// MARK: - Stateful Lifecycle Handlers
+
+/// LifecycleHandler for starting stateful tasks. The state can then be fed into a LifecycleShutdownHandler
+public struct LifecycleStartHandler<State> {
+    public typealias Handler = (@escaping (Result<State, Error>) -> Void) -> Void
```
weissi, line 103: I would not use `typealias` for functions, especially not in public API. The reason is that in Xcode you don't get useful completions. Ie the user won't know what type of function they have to implement. They'll have to Cmd-click `Handler`, then copy and paste the type. It's harder to write (for you) without the typealias but it's much easier for users to work with it.

tomerd: good idea. addressed

tomerd:
@yim-lee @ktoso @weissi @rauhul

yim-lee (approved):

tomerd:

tomerd:

tomerd:

yim-lee (approved):

README.md
```diff
@@ -214,6 +214,32 @@ In more complex cases, when `Signal`-trapping-based shutdown is not appropriate,
 
 `shutdown` is an asynchronous operation. Errors will be logged and bubbled up to the provided completion handler.
 
+### Stateful handlers
+
+In some cases it is useful to have the Start handlers return a state that can be passed on to the Shutdown handlers for shutdown.
+For example, when establishing some sort of a connection that needs to be closed at shutdown.
+
+```swift
+struct Foo {
```
ktoso, line 223: Or if it conforms to `LifecycleStartHandler` one could just pass it as?

```
 lifecycle.registerStateful(
     label: "foo",
     handler: foo
 )
```

or not possible?

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -82,15 +82,97 @@ public struct LifecycleHandler {
...
+
+/// LifecycleHandler for starting stateful tasks. The state can then be fed into a LifecycleShutdownHandler
+public struct LifecycleStartHandler<State> {
+    private let underlying: (@escaping (Result<State, Error>) -> Void) -> Void
+
+    /// Initialize a `LifecycleHandler` based on a completion handler.
+    ///
+    /// - parameters:
+    ///    - callback: the underlying completion handler
+    public init(_ handler: @escaping (@escaping (Result<State, Error>) -> Void) -> Void) {
```
ktoso, line 107: is that double escaping needed?

README.md
ktoso: Ah i see... you only surface the new APIs really, not forcing people to impl anything. That sounds good 👍

ktoso (approved):
Awesome, just what I'd need/love :-) 

I think the API is expressed also very nicely, does not differ much from the existing one etc.

LGTM!

Issue #35, opened 2020-04-10
express backtrace as a LifecycleTask


tomerd:
Should installing backtrace be expressed as a LifecycleTask which we automatically add when it's enabled

hassila:
@tomerd should be sorted out with https://github.com/swift-server/swift-service-lifecycle/pull/87

hassila:
I think this case can be closed after the discussion in #87 ?

tomerd:
yes, I think backtraces are special enough so they should be installed as early as possible - at the init imo

PR #87, opened 2021-03-13
Express backtrace as a LifecycleTask

hassila:
**Motivation**
Refactor to use LifecycleTask to start the Backtrace according to https://github.com/swift-server/swift-service-lifecycle/issues/35

**Changes**
Deferred installation of Backtrace from init to instead use a LifecycleTask.

**Results**
Backtrace is started slightly later in the startup chain (e.g. after setting up signal handlers) - should be ok though, it will be the first installed task.

### DIFF
--- a/.xcodesamplecode.plist
+++ b/.xcodesamplecode.plist
@@ -1,5 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
-<plist version="1.0">
-<array/>
-</plist>
--- a/Sources/Lifecycle/Lifecycle.swift
+++ b/Sources/Lifecycle/Lifecycle.swift
@@ -504,8 +504,8 @@ extension LifecycleTasksContainer {
     ///    - label: label of the item, useful for debugging.
     ///    - start: `Handler` to perform the startup.
     ///    - shutdown: `Handler` to perform the shutdown.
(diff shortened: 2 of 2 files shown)


ktoso:
@swift-server-bot add to whitelist

tomerd:
thanks for the PR @hassila! I have mixed feeling about this idea (the implementation is great!) -- having the backtraces installed as soon as possible is desired since event the initial setup may cause a crash and without the backtraces installed we would not get backtraces on those

tomerd:

hassila:
Yeah, I also thought about that - but figured it would be ok in terms of tradeoff as we will be the absolutely first handler installed - the window of "opportunity" for things to fail is a very small initial path while we are setting up - and then we would have died very early indeed - in practice a release worthy version of the library shouldn't keel over there (as a user I'd be ok with resorting to running it in the debugger if that ever would happen in practice - I'd probably see if usable core files were an option for such cases).

hassila:
Or to put it another way - I would expect 99.9999% of all actual crashes happen later due to users of the framework doing something bad... (famous last words, I know)

hassila:
@tomerd In the same vein, I would propose to move the installation of signal handles to also be LifecycleTasks and install them directly after backtrace in a similar way. That would also decrease the amount of code being run before backtrace is installed.

tomerd:
setting up the shutdown signal handler sounds good first step, backtraces are still better as a discrete first action imo

hassila:
That's absolutely fine of course, closing this one. Opened https://github.com/swift-server/swift-service-lifecycle/pull/88 for the signal handler installation for consideration.

PR #88, opened 2021-03-14
Install signal handlers as LifecycleTasks

hassila:
Motivation:

Small code cleanup, reusing LifecycleTask as the shutdown hooks are
to be installed just at the beginning of that chain of events anyway.
Discussed previously in context of https://github.com/swift-server/swift-service-lifecycle/pull/87

Modifications:

Moved signal handler installation to LifecycleTask instead of direct calls.

Result:

Slightly simpler code.

### DIFF
--- a/Sources/Lifecycle/Lifecycle.swift
+++ b/Sources/Lifecycle/Lifecycle.swift
@@ -201,9 +201,11 @@ public struct ServiceLifecycle {
         self.underlying = ComponentLifecycle(label: self.configuration.label, logger: self.configuration.logger)
         // setup backtraces as soon as possible, so if we crash during setup we get a backtrace
         self.installBacktrace()
-        self.register(label: "Shutdown hooks",
-                      start: .sync(self.setupShutdownHook),
-                      shutdown: .none)
+        if self.configuration.shutdownSignal != nil {
+            self.register(label: "Shutdown hooks",
+                          start: .sync(self.setupShutdownHook),
+                          shutdown: .none)
+        }
     }
 
     /// Starts the provided `LifecycleTask` array.
--- a/Sources/Lifecycle/Lifecycle.swift
+++ b/Sources/Lifecycle/Lifecycle.swift
@@ -183,7 +183,7 @@ public struct LifecycleShutdownHandler<State> {
 ///  By default, also install shutdown hooks based on `Signal` and backtraces.
 public struct ServiceLifecycle {
     private static let backtracesInstalled = AtomicBoolean(false)
-    private static let signalHandlerInstalled = AtomicBoolean(false)
+    private static let shutdownHooksInstalled = AtomicBoolean(false)
 
     private let configuration: Configuration
 
@@ -202,7 +202,7 @@ public struct ServiceLifecycle {
         self.underlying = ComponentLifecycle(label: self.configuration.label, logger: self.configuration.logger)
         // setup backtraces as soon as possible, so if we crash during setup we get a backtrace
         self.installBacktrace()
-        self.installSignalHandler()
+        self.installShutdownHooks()
     }
 
     /// Starts the provided `LifecycleTask` array.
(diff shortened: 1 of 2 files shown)


Sources/Lifecycle/Lifecycle.swift
```diff
@@ -201,6 +201,11 @@ public struct ServiceLifecycle {
         self.underlying = ComponentLifecycle(label: self.configuration.label, logger: self.configuration.logger)
         // setup backtraces as soon as possible, so if we crash during setup we get a backtrace
         self.installBacktrace()
+        if self.configuration.shutdownSignal != nil {
+            self.register(label: "Shutdown hooks",
+                          start: .sync(self.setupShutdownHook),
+                          shutdown: .none)
+        }
```
ktoso, line 208: should this rather be `self.installShutdownSignalHooks`? 

It'd be the same style as install backtrace and avoid having to much logic in the init function itself

tomerd: this is great @hassila!

+1 to abstract this to an  function "installShutdownSignalHook" function

hassila: I did a slight reiteration along those lines while also adding support for proper cancellation fo all dispatch signal sources during shutdown, please have a look again.

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -201,6 +203,7 @@ public struct ServiceLifecycle {
         self.underlying = ComponentLifecycle(label: self.configuration.label, logger: self.configuration.logger)
         // setup backtraces as soon as possible, so if we crash during setup we get a backtrace
         self.installBacktrace()
+        self.installSignalhandler()
```
tomerd, line 206: `installSignalHandler`

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -247,17 +248,28 @@ public struct ServiceLifecycle {
         }
     }
 
-    private func setupShutdownHook() {
+    private func installSignalhandler() {
```
tomerd, line 251: `installSignalHandler`

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -183,6 +183,8 @@ public struct LifecycleShutdownHandler<State> {
 ///  By default, also install shutdown hooks based on `Signal` and backtraces.
 public struct ServiceLifecycle {
     private static let backtracesInstalled = AtomicBoolean(false)
+    private static let signalHandlerInstalled = AtomicBoolean(false)
+    private static var signalHandlerSources = [DispatchSourceSignal]()
```
tomerd, line 187: this is not thread safe, also why static?

tomerd: I am confused on why we need this in the firs place, aren't the shutdown hook already cancel cleanly?

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -247,17 +248,28 @@ public struct ServiceLifecycle {
...
+            }, cancelAfterTrap: true))
+        }
+    }
+
+    private func deinstallShutdownHook() {
+        for source in ServiceLifecycle.signalHandlerSources {
+            source.cancel()
         }
+        ServiceLifecycle.signalHandlerSources = []
     }
```
tomerd, line 273: see above, not sure why this is required

Sources/Lifecycle/Lifecycle.swift
hassila: Static as I could find no way to get it mutating properly for the instance - if making installSignalHandler mutating, I could not be able to register it as a LifecycleTask.

hassila: signalHandlerSources is only written to once by the creating thread (using the CAS (similar to the backtrace installation)) - the actual signal register/shutdown is done in the context of the LifecycleTask and would presumably be ok?

Sources/Lifecycle/Lifecycle.swift
hassila: It was intended to release any references to the signal dispatch sources. Maybe it was a bit overkill and I can revert to the previous cancellation approach which would only disable the signal dispatch source for the signal that was triggered.

Sources/Lifecycle/Lifecycle.swift
hassila: reverted to previos simpler approach.

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -247,9 +247,16 @@ public struct ServiceLifecycle {
         }
     }
 
-    private func setupShutdownHook() {
+    private func installSignalHandler() {
```
tomerd, line 250: nit: installShutdownHooks

tomerd (approved):

tomerd (approved):

Sources/Lifecycle/Lifecycle.swift
hassila: d341896

hassila (commented):
LGTM

PR #90, opened 2021-03-17
Revert "Install signal handlers as LifecycleTasks"

tomerd:
motivation: start handlers should not actually be part of the initialization process, as it breaks composition between ServiceLifecycles

### DIFF
--- a/Sources/Lifecycle/Lifecycle.swift
+++ b/Sources/Lifecycle/Lifecycle.swift
@@ -183,7 +183,6 @@ public struct LifecycleShutdownHandler<State> {
 ///  By default, also install shutdown hooks based on `Signal` and backtraces.
 public struct ServiceLifecycle {
     private static let backtracesInstalled = AtomicBoolean(false)
-    private static let shutdownHooksInstalled = AtomicBoolean(false)
 
     private let configuration: Configuration
 
@@ -202,7 +201,6 @@ public struct ServiceLifecycle {
         self.underlying = ComponentLifecycle(label: self.configuration.label, logger: self.configuration.logger)
         // setup backtraces as soon as possible, so if we crash during setup we get a backtrace
         self.installBacktrace()
-        self.installShutdownHooks()
     }
 
     /// Starts the provided `LifecycleTask` array.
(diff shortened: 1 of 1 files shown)


fabianfett (approved):

tomerd:
approved by @fabianfett

tomerd:
hi @hassila, we needed to revert this as it breaks composability of ServiceLifeycle, sorry I did not notice that in the review!

hassila:
@tomerd ok, not sure I fully understand the problem, but thanks for notification.

fabianfett:
@hassila The problem with the approach taken is, if users create multiple `ServiceLifecycle`s there is no way for them to know, which of those `ServiceLifecycle`s was able to register for signals. They might be assuming all of them listen for a signal. This becomes very in-transparent. For this reason we are considering to make the logic as explicit as possible, by crashing as soon as another ServiceLifecycle has been created: See https://github.com/swift-server/swift-service-lifecycle/pull/92

hassila:
All of them would be listening, dispatch sources support delivering to multiple signal dispatch sources, no? Then of course, they would all initiate shutdowns...

hassila:
(although agree with #92, it seems a bit weird to have multiple instances of ServiceLifecycle)

fabianfett:
> All of them would be listening, dispatch sources support delivering to multiple signal dispatch sources, no?

Yes. But #88 changed the logic that only the first `ServiceLifecycle` would register for those signals. `shutdownHooksInstalled` was a static variable.

hassila:
Ok, so just removing shutdownHooksInstalled and the check for that and things would be ok? Those were only added as I understood @tomerd wanted it analogue with the backtrace installation, but I may have misunderstood.

fabianfett:
> Ok, so just removing shutdownHooksInstalled and the check for that and things would be ok?

Kinda, the problem then would be that the distinction between `ServiceLifecycle` and `ComponentLifecycle` would be quite hard to explain anyways. I think the new precondition introduced in #92, makes sure developers use the right tool at the right time.

PR #92, opened 2021-03-17
RFC: validate that only a single instance of ServiceLifecycle is created per process

tomerd:
motivation: having multiple instance of `ServiceLifecycle` generally points to a mistake, which we can help validate

changes: add check in **init** to validate single instance

alternatively, we may be less rigid, we can move this check to `start` so multiple instance may be created, but only one started - I think I prefer this option

### DIFF
--- a/Sources/Lifecycle/Lifecycle.swift
+++ b/Sources/Lifecycle/Lifecycle.swift
@@ -180,9 +180,12 @@ public struct LifecycleShutdownHandler<State> {
 // MARK: - ServiceLifecycle
 
 /// `ServiceLifecycle` provides a basic mechanism to cleanly startup and shutdown the application, freeing resources in order before exiting.
-///  By default, also install shutdown hooks based on `Signal` and backtraces.
+/// By default, also install shutdown hooks based on `Signal` and backtraces.
+/// `ServiceLifecycle` is designed to be used by executables/applications, and as such there should only a single copy of `ServiceLifecycle` loaded into the memory space
+/// Creating multiple copies of `ServiceLifecycle` will lead to a crash.
+/// Libraries should always use `ComponentLifecycle` for composition.
 public struct ServiceLifecycle {
-    private static let backtracesInstalled = AtomicBoolean(false)
+    private static let initialized = AtomicBoolean(false)
 
     private let configuration: Configuration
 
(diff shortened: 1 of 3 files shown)


Sources/Lifecycle/Lifecycle.swift
```diff
@@ -197,6 +197,18 @@ public struct ServiceLifecycle {
...
+    }
+
+    // for testing
+    internal static func makeForTesting(configuration: Configuration = .init()) -> ServiceLifecycle {
+        return ServiceLifecycle(validateSingleInstance: false, configuration: configuration)
+    }
+
+    private init(validateSingleInstance: Bool = true, configuration: Configuration) {
+        if validateSingleInstance, !ServiceLifecycle.initialized.compareAndSwap(expected: false, desired: true) {
+            fatalError("only a single instance of ServiceLifecycle per process is allowed")
```
fabianfett, line 210: Wanna add a file and line where the bad second configuration is created?

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -197,6 +197,18 @@ public struct ServiceLifecycle {
     /// - parameters:
     ///    - configuration: Defines lifecycle `Configuration`
     public init(configuration: Configuration = .init()) {
```
fabianfett, line 199: I think, we should update the header comment. It should say something like, there should only be one ServiceLifecycle per executable. For composition always use `ComponentLifecycle`. Creating a second `ServiceLifecycle` will lead to a crash.

fabianfett (approved):

yim-lee (approved):

fabianfett (commented):
Two small issues re documentation.

RE: Crash in `init` vs `start`. I can't see the benefit of crashing only if start has been called on multiple `ServiceLifecycle` instances. Generally I prefer to fail as early as possible.

fabianfett (approved):
Thanks! I really like this. 🚀

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -180,9 +180,12 @@ public struct LifecycleShutdownHandler<State> {
 // MARK: - ServiceLifecycle
 
 /// `ServiceLifecycle` provides a basic mechanism to cleanly startup and shutdown the application, freeing resources in order before exiting.
-///  By default, also install shutdown hooks based on `Signal` and backtraces.
+/// By default, also install shutdown hooks based on `Signal` and backtraces.
```
ktoso, line 183:
```suggestion
///
/// By default, also install shutdown hooks based on `Signal` and backtraces.
```

give it a bit more breathing space from the first sentence

ktoso (approved):
I guess this can be fine, usually I don't love such checks but indeed with lifecycle it seems appropriate .

fabianfett:
@tomerd Wanna go forward with this PR?

FranzBusch:
Going to close this now since our new APIs don't have that restriction.

PR #93, opened 2021-03-18
add lifecycle metrics

tomerd:
motivation: startup/shutdown metrics are important for real-life services, for example spike in start metrics indicates a crash-loop

changes:
* add start and shutdown counters to ComponentLifecycle
* add start and shutdown timers to report duration of startup and shutdown operations

### DIFF
--- a/Sources/Lifecycle/Lifecycle.swift
+++ b/Sources/Lifecycle/Lifecycle.swift
@@ -18,9 +18,9 @@ import Darwin
 import Glibc
 #endif
 import Backtrace
+import CoreMetrics
 import Dispatch
 import Logging
-import Metrics
 
 // MARK: - LifecycleTask
 
@@ -473,9 +473,12 @@ public class ComponentLifecycle: LifecycleTask {
             guard case .idle = self.state else {
                 preconditionFailure("invalid state, \(self.state)")
             }
-            self.log("starting")
             self.state = .starting(queue)
         }
+
+        self.log("starting")
+        Counter(label: "\(self.label).lifecycle.start").increment()
+
         if tasks.count == 0 {
             self.log(level: .notice, "no tasks provided")
         }
@@ -517,7 +520,9 @@ public class ComponentLifecycle: LifecycleTask {
             return callback(index, nil)
         }
         self.log("starting tasks [\(tasks[index].label)]")
+        let startTime = DispatchTime.now()
         start { error in
+            Timer(label: "\(self.label).\(tasks[index].label).lifecycle.start").recordNanoseconds(DispatchTime.now().uptimeNanoseconds - startTime.uptimeNanoseconds)
             if let error = error {
                 self.log(level: .error, "failed to start [\(tasks[index].label)]: \(error)")
                 return callback(index, error)
@@ -532,9 +537,12 @@ public class ComponentLifecycle: LifecycleTask {
 
     private func _shutdown(on queue: DispatchQueue, tasks: [LifecycleTask], callback: @escaping () -> Void) {
         self.stateLock.withLock {
-            log("shutting down")
             self.state = .shuttingDown(queue)
         }
+
+        self.log("shutting down")
+        Counter(label: "\(self.label).lifecycle.shutdown").increment()
+
         self.shutdownTask(on: queue, tasks: tasks.reversed(), index: 0, errors: nil) { errors in
             self.stateLock.withLock {
                 guard case .shuttingDown = self.state else {
@@ -555,8 +563,11 @@ public class ComponentLifecycle: LifecycleTask {
         if index >= tasks.count {
             return callback(errors)
         }
+
         self.log("stopping tasks [\(tasks[index].label)]")
+        let startTime = DispatchTime.now()
         shutdown { error in
+            Timer(label: "\(self.label).\(tasks[index].label).lifecycle.shutdown").recordNanoseconds(DispatchTime.now().uptimeNanoseconds - startTime.uptimeNanoseconds)
             var errors = errors
             if let error = error {
                 if errors == nil {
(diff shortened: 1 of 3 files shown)


Sources/Lifecycle/Lifecycle.swift
```diff
@@ -473,9 +473,12 @@ public class ComponentLifecycle: LifecycleTask {
             guard case .idle = self.state else {
                 preconditionFailure("invalid state, \(self.state)")
             }
-            self.log("starting")
             self.state = .starting(queue)
         }
+
+        self.log("starting")
+        Counter(label: "\(self.label).lifecycle.start").increment()
```
ktoso, line 480: is label risky i.e. containing spaces etc?

yim-lee (approved):

Sources/Lifecycle/Lifecycle.swift
tomerd: hmm good question, do you think we should normalize it?

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -18,9 +18,9 @@ import Darwin
 import Glibc
 #endif
 import Backtrace
+import CoreMetrics
```
avolokhov, line 21: is it intentional? AFAIR libraries shouldn't normally depend on CoreMetrics unless implementing a MetricFactory.

Sources/Lifecycle/Lifecycle.swift
ktoso: Was just thinking about it but I also just remembered that last time we said that the libs should sanitize, so we do e.g. in the prometheus one https://github.com/MrLotU/SwiftPrometheus/blob/f4d4adfeb1178e250a0ea427017fb4d1cc877de3/Sources/Prometheus/PrometheusMetrics.swift#L112 🤔 


I guess it's fine 👍

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -517,7 +520,9 @@ public class ComponentLifecycle: LifecycleTask {
             return callback(index, nil)
         }
         self.log("starting tasks [\(tasks[index].label)]")
+        let startTime = DispatchTime.now()
         start { error in
+            Timer(label: "\(self.label).\(tasks[index].label).lifecycle.start").recordNanoseconds(DispatchTime.now().uptimeNanoseconds - startTime.uptimeNanoseconds)
```
ktoso, line 525: Could it be worth making some struct/enum with the labels such that it is easier to look in one place to see "aha, these are all the metric labels this lib exports"?

I do this in some other project, it is easier to know what metrics are exposed then.
Sure not all that much here but it's a nice pattern

avolokhov (commented):
Graphite/statsd and Prometheus have different naming schemes (dot notation vs snake_case).
For my own education, is it a general direction to use dot notation?

Sources/Lifecycle/Lifecycle.swift
ktoso: We have `    func recordInterval(since: DispatchTime, now: DispatchTime = .now()) {` for this type of nanosecond recording, consider using it

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -555,8 +563,11 @@ public class ComponentLifecycle: LifecycleTask {
         if index >= tasks.count {
             return callback(errors)
         }
+
         self.log("stopping tasks [\(tasks[index].label)]")
+        let startTime = DispatchTime.now()
         shutdown { error in
+            Timer(label: "\(self.label).\(tasks[index].label).lifecycle.shutdown").recordNanoseconds(DispatchTime.now().uptimeNanoseconds - startTime.uptimeNanoseconds)
```
ktoso, line 570: Consider using `recordinternal(since: startTime)`

ktoso (approved):

Sources/Lifecycle/Lifecycle.swift
tomerd: CoreMetrics is useful when you do not want to pul in Foundation

Sources/Lifecycle/Lifecycle.swift
tomerd: recordInterval is not define in the core library afaict

tomerd:
> Graphite/statsd and Prometheus have different naming schemes (dot notation vs snake_case).
For my own education, is it a general direction to use dot notation?

I do not believe there is a standard for this, and for this reason our MetricsAPI is also is not opinionated about this: as you pointed out different system (Graphite, Prometheus) have their own conventions and constraints so it must be the "job" of the "backend" library to sanitize the labels to whatever its sensitive to, for example Prometheus is sensitive to dot notation, so its the "job" of the Prometheus "backend" library to convert dots into underscores or whatever.

Sources/Lifecycle/Lifecycle.swift
ktoso: We have it here https://github.com/apple/swift-metrics/blob/44e8bfc7f5a0686e21d5d21a7ea5454e74ed3449/Sources/Metrics/Metrics.swift#L42-L44

maybe it's not released yet? 🤔

ktoso: I see but it's not in CoreMetrics hm

ktoso:
Yeah agreed on the sanitization, sounds good. LGTM

Sources/Lifecycle/Lifecycle.swift
tomerd: ah yes... I prefer not to introduce dependency on Foundation  for low level libs if we can get away with it

ktoso: Yeah though this could live in core I guess, I'll ticketify -- it's just Dispatch dep, not foundation hm

PR #95, opened 2021-05-02
Remove [Lifecycle] prefix from log messages

slashmo:
**Motivation:**

Since Swift Log 1.3.0, log handlers can include the source of a particular message,
rendering the [Lifecycle] prefix unnecessary.

**Modifications:**

Removed the [Lifecycle] prefix from all log messages by removing the underlying
log helper and calling out to the Logger itself.

**Result:**

Log messages written by a log handler which includes the location won't log "Lifecycle" twice.

**Related:**

- [Logging (#135): "add source for log messages"](https://github.com/apple/swift-log/pull/135)
- [Logging (#187): "use #fileID for Swift 5.3+"](https://github.com/apple/swift-log/pull/187)
- [Logging (#189): "Include source in StreamLogHandler output"](https://github.com/apple/swift-log/pull/189)

### DIFF
--- a/Sources/Lifecycle/Lifecycle.swift
+++ b/Sources/Lifecycle/Lifecycle.swift
@@ -258,7 +258,7 @@ public struct ServiceLifecycle {
     }
 
     private func log(_ message: String) {
-        self.underlying.log(message)
+        self.underlying.logger.info("\(message)")
     }
 }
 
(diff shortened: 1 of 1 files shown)


slashmo:
I created this as a draft for now as I'm not sure how to handle backward compatibility here. Any feedback is highly appreciated!

tomerd:

tomerd:
personally, I think this is fine without backward compatibility. however if there is a consensus it is important you can make it into a configuration in 'Configuration' so folks can opt-in if they still want it

tomerd:
@fabianfett @ktoso @yim-lee opinions?

fabianfett (approved):

fabianfett:
I don't think we need backwards compatibility for this lib. I support this change!

ktoso (approved):
LGTM :-)

I'm partial to this change as I suggested/requested @slashmo made the PR but yes, I think that's absolutely the right thing to do 👍 

No concerns about compatibility, we don't guarantee stability of log messages like that.

Issue #94, opened 2021-03-30
adopt async/await


tomerd:

kirilltitov:
BTW this simple extension worked for me perfectly :)

```swift
import _Concurrency
public extension LifecycleHandler {
    init(_ handler: @escaping () async throws -> Void) {
        self = LifecycleHandler { callback in
            // TODO: change to `spawnDetached` when structured concurrency proposal is accepted
            Task.runDetached {
                do {
                    try await handler()
                    callback(nil)
                } catch {
                    callback(error)
                }
            }
        }
    }
}

```

tomerd:
closed via https://github.com/swift-server/swift-service-lifecycle/pull/99

PR #99, opened 2021-05-07
add handlers that support async functions

tomerd:
motivation: async/await is coming in 5.5, take first steps to support it

changes:
* introduce handler initializers that takes async functions (@escaping () async throws -> Void, @escaping () async throws -> State)
* add tests

### DIFF
--- a/Sources/Lifecycle/Lifecycle.swift
+++ b/Sources/Lifecycle/Lifecycle.swift
@@ -12,6 +12,10 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if compiler(>=5.5)
+import _Concurrency
+#endif
+
 #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
 import Darwin
 #else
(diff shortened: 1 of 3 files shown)


tomerd:
fixes #94

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -12,6 +12,10 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if compiler(>=5.5)
+import _Concurrency
```
yim-lee, line 16: Is this import required?

yim-lee (approved):

Sources/Lifecycle/Lifecycle.swift
ktoso: It'll be gone eventually (it should be already but seems something went wrong)

ktoso (approved):
Ah so just the small function additions for now, LGTM :)

PR #102, opened 2021-06-09
adjust to updated concurrency API

tomerd:
motivation: `detach` was replaced with `Task` and `Task.detached`

changes: use `Task {}` instead of `detach {}`


yim-lee (approved):

ktoso:
lgtm 👍

PR #104, opened 2021-08-19
fix repeated shutdown issue on linux

tomerd:
motivation: repeatedly calling shutdown in linux was broken due to multiple calls to singal() which work on darwin but not linux

changes:
* conditionalize the call to signal
* add test


yim-lee (approved):

tomerd:
cc @oc243

oc243:
Thanks so much @tomerd

PR #107, opened 2021-09-02
when shutdownIfNotStarted is set to false, dont shutdown items that failed to start

tomerd:
motivation: improve confusing behavior

changes:
* change startTask to return an array of started tasks instead of an index
* only shutdown successfully started tasks or ones that have explicitly set shutdownIfNotStarted to true
* add and adjust tests

### DIFF
--- a/Sources/Lifecycle/Lifecycle.swift
+++ b/Sources/Lifecycle/Lifecycle.swift
@@ -576,11 +576,13 @@ public class ComponentLifecycle: LifecycleTask {
             case .shuttingDown:
                 self.stateLock.unlock()
                 // shutdown was called while starting, or start failed, shutdown what we can
-                let stoppable: [LifecycleTask]
-                if started < tasks.count {
-                    stoppable = tasks.enumerated().filter { $0.offset <= started || $0.element.shutdownIfNotStarted }.map { $0.element }
-                } else {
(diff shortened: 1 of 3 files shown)


tomerd:
jazzy issues failing CI, need #108 first

Sources/Lifecycle/Lifecycle.swift
```diff
@@ -590,25 +592,27 @@ public class ComponentLifecycle: LifecycleTask {
...
             if let error = error {
                 self.logger.error("failed to start [\(tasks[index].label)]: \(error)")
-                return callback(index, error)
+                let started = Array(tasks.prefix(index))
+                return callback(started, error)
             }
             // shutdown called while starting
             if case .shuttingDown = self.stateLock.withLock({ self.state }) {
-                return callback(index, nil)
+                let started = index < tasks.count ? Array(tasks.prefix(index + 1)) : tasks
```
yim-lee, line 614: Is the task at `index` considered started?

tomerd: this is a non-error case, meaning the task was started successfully. do you see something else?

tomerd:
@weissi do you want to test this one to make sure it addressed your concern?

yim-lee (approved):

PR #109, opened 2021-09-02
shutdown any task that have shutdownIfNotStarted set to true when shutting down a never-started-lifecycle

tomerd:
motivation: more reliable shutdown in edge cases where lifecycle never gets started

changes: when shutting down an idle lifecycle, try to shutdown any tasks that have shutdownIfNotStarted set to true

### DIFF
--- a/Sources/Lifecycle/Lifecycle.swift
+++ b/Sources/Lifecycle/Lifecycle.swift
@@ -507,11 +507,17 @@ public class ComponentLifecycle: LifecycleTask {
 
         self.stateLock.lock()
         switch self.state {
-        case .idle:
+        case .idle(let tasks) where tasks.isEmpty:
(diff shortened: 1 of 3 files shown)


tomerd:
jazzy issues failing CI, need #108 first

yim-lee (approved):

fabianfett (approved):
LGTM. Maybe we should explicitly call out this behavior in the docs. IMHO it is quite important, if you use MultiThreadedEventLoopGroup, AsyncHTTPClient or Soto.

tomerd:
> Maybe we should explicitly call out this behavior in the docs. IMHO it is quite important, if you use MultiThreadedEventLoopGroup, AsyncHTTPClient or Soto.

thats a good idea. was planning to rework docs in a following PR anyways, so will do then

Issue #110, opened 2021-09-16
Fails to compile Xcode 13 RC (note: needs macOS release)


joshuawright11:
Hey folks, trying to build with Xcode 13 RC.

### Steps to reproduce
`git clone https://github.com/swift-server/swift-service-lifecycle.git`
`cd swift-service-lifecycle`
`swift build`

```
.../swift-service-lifecycle/Sources/Lifecycle/Lifecycle.swift:16:8: error: no such module '_Concurrency'
import _Concurrency
```

And `swift --version` is
```
swift-driver version: 1.26.9 Apple Swift version 5.5 (swiftlang-1300.0.31.1 clang-1300.0.29.1)
Target: x86_64-apple-macosx11.0
```

~Assume the fix is just to delete the `import _Concurrency`?~ (didn't actually fix it, as discussed below)

ktoso:
Yeah, the import is not necessary anymore, do you want to send in a PR?

ktoso:
Ok sorry I know what's going on now.

This is a weird situation in time where the RC of Xcode assumes that concurrency is provided by the OS -- the Betas did not assume that but this RC does. But the OS with Concurrency has not been released yet, so you can't use the RC Xcode yet.

This will just work once the next macOS is released, it will have the next SDK with concurrency and things will just work.

joshuawright11:
Gotcha, okay. How does that work on linux I wonder? Is concurrency provided on the linux swift toolchains when >=5.5?

ktoso:
On linux everything is included in the toolchains yes.

joshuawright11:
👍 so just to be clear there is no way to use the new async await stuff with Xcode 13 RC until I've got macOS Monterey is installed?

(to use it I'd have to wait for Monterey to drop or use a prior Xcode 13 Beta)

ktoso:
That seems to be the case AFAICS

joshuawright11:
Since the official Xcode 13 is out and this still seems to prevent compilation for macOS targets other than the unreleased Monterrey, would it make more sense to instead of using

```swift
#if compiler(>=5.5)
```

use this?

```swift
#if canImport(_Concurrency)
```

Since technically using concurrency is dependent on the availability of the concurrency library rather than swift 5.5? And even after Monterrey is released compilation would still fail on older macOS targets. 

Perhaps it would also need to check for 5.5 as well;

```swift 
#if compiler(>=5.5) && canImport(_Concurrency)
```

[...]

PR #111, opened 2021-09-21
Switch up compiler directives for concurrency

joshuawright11:
This may be misguided, see comments in #110

### DIFF
--- a/Sources/Lifecycle/Lifecycle.swift
+++ b/Sources/Lifecycle/Lifecycle.swift
@@ -12,10 +12,6 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if compiler(>=5.5) && canImport(_Concurrency)
-import _Concurrency
-#endif
-
 #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
 import Darwin
 #else
--- a/Sources/Lifecycle/Lifecycle.swift
+++ b/Sources/Lifecycle/Lifecycle.swift
@@ -12,7 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if compiler(>=5.5)
+#if canImport(_Concurrency)
 import _Concurrency
 #endif
 
@@ -99,7 +99,7 @@ public struct LifecycleHandler {
     }
 }
 
-#if compiler(>=5.5)
+#if canImport(_Concurrency)
 @available(macOS 12.0, *)
 extension LifecycleHandler {
     public init(_ handler: @escaping () async throws -> Void) {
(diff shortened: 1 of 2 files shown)


tomerd:
@ktoso is this the right approach?

joshuawright11:
Seeing as CI is failing, I assume the tests are all running on an OS where `_Concurrency` is available.  Added `compiler(>=5.5)` to the conditionals.

ktoso (commented):
This is _partially_ correct.

- We should _remove_ any mention of `import _Concurrency` at all; completely. This module is imported automatically when it exists. It should not be imported explicitly - it just adds more noise to the project.

- The `#if` guard IS correct though for guarding any code _actually using async/Concurrency_, we can't just detect the swift version but have to guard with the concurrency import availability as well.

Would you mind changing the PR to remove the imports? Then it's LGTM :)

adam-fowler:
Once Monterey is out should we still be checking for existence of _Concurrency. Is this a temporary fix until that point?

ktoso:
For now assume it'll be around.

joshuawright11:
Great, thanks @ktoso. Just pushed.

ktoso:
@swift-ci please smoke test

ktoso:

tomerd:
thanks @joshuawright11 something does not seem to be correct with the conditions give the CI failures in > 5.2 < 5.5 version

ktoso:
Something is indeed weird here... I was checking with the compiler team if these are the guards we should use and we think those are right 🤔

tomerd:
I had trouble composing them in the past, hence the guards

joshuawright11:
Yeah it does seem like something odd is going on, I poked around with the complex #if statements on my machine and it worked fine; though my machine is also 5.5 so perhaps it's an issue with older swift versions.

Alas, `guard #canImport(_Concurrency)` gave an `error: use of unknown directive '#canImport`, so I split up the if statement further to:

```swift
#if compiler(<5.2)
return
#elseif compiler(<5.5)
throw XCTSkip()
#elseif !canImport(_Concurrency)
throw XCTSkip()
#else
...
#endif
```

Let's see if it likes that.

tomerd:

tomerd:
> Yeah it does seem like something odd is going on, I poked around with the complex #if statements on my machine and it worked fine; though my machine is also 5.5 so perhaps it's an issue with older swift versions.

fyi you can run the tests we run in CI locally using docker, see the docker-compose files

tomerd (approved):

tomerd:
nightly CI failure expected due to @sendable changes

PR #114, opened 2021-09-27
Use swift-atomics from v1.0.0

adam-fowler:
Now swift-atomics has hit v1.0.0 we should use it

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -19,7 +19,7 @@ let package = Package(
 
 #if compiler(>=5.2)
 package.dependencies += [
-    .package(url: "https://github.com/apple/swift-atomics.git", .exact("0.0.3")), // exact since < 1.0
+    .package(url: "https://github.com/apple/swift-atomics.git", from: "1.0.0"), // exact since < 1.0
 ]
 package.targets += [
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -19,7 +19,7 @@ let package = Package(
 
 #if compiler(>=5.2)
 package.dependencies += [
-    .package(url: "https://github.com/apple/swift-atomics.git", .exact("0.0.3")), // exact since < 1.0
+    .package(url: "https://github.com/apple/swift-atomics.git", from: "1.0.0"), // exact since < 1.0
```
tomerd, line 22:
```suggestion
    .package(url: "https://github.com/apple/swift-atomics.git", from: "1.0.0"),
```

tomerd:

tomerd:
welp, look like atomics now requires 5.3?

adam-fowler:
> welp, look like atomics now requires 5.3?

The commit message for this references https://github.com/apple/swift-atomics/issues/22. Issue seems to be about supporting 5.2 and above but instead they updated to 5.3. :-/

tomerd:
> The commit message for this references apple/swift-atomics#22. Issue seems to be about supporting 5.2 and above but instead they updated to 5.3. :-/

well the code as fallback for using the "old" atomics, we just need to move the check to < 5.3

PR #117, opened 2021-12-16
update doc generation script

tomerd:
motivation: source-kitten has changed how its invoked

changes:
* invoke source-kitten with the correct flags
* make location of source-kitten source checkout more robust
* udpate module list


yim-lee (approved):

ktoso:
LGTM

PR #130, opened 2023-03-10
Overhaul service lifecycle for Structured Concurrency

FranzBusch:
# Motivation
Since the release of Swift Concurrency the server ecosystem has adopted `async`/`await` in a lot of APIs making the user experience amazing. Another thing that Swift Concurrency introduced is Structured Concurrency which allows to create a task tree with parent to child relations. Structured Concurrency has many benefits such as automatic cancellation and task local propagation. The current state of this library predates the introduction of Swift Concurrency and is not inline anymore with how applications are structured that want to leverage Concurrency.

# Modification
This PR overhauls the implementation of service lifecycle completely. The reason for this is that the current implementation is very focused on how NIO based applications work before the introduction of Concurrency. Furthermore, Structured Concurrency is actually replacing part of the current functionality by providing new scoping mechanisms. The overhauled implementation provides two primitive types. Anew `Service` protocol and a `ServiceRunner`. The former is providing a clear API which services have to conform to so that they can be run by the `ServiceRunner`.

An example usage of the types looks like this

```swift
actor FooService: Service {
    func run() async throws {}
}
actor BarService: Service {
    func run() async throws {}
}

let fooService = FooService()
let barService = BarService()

let runner = ServiceRunner(
   services: [fooService, barService],
   configuration: .init(gracefulShutdownSignals: [.sigterm]),
   logger: logger
)

try await runner.run()
```

# Result
We now have a service lifecycle library that integrates nicely with Structured Concurrency. Its value add is that it solves to complex setup of the task group with the signal handling. Furthermore, it provides a currency type `Service` that can be passed to inject services.


Sources/ServiceLifecycle/CancellableContinuation.swift
```diff
@@ -0,0 +1,37 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+/// A simple actor that provides a method that suspends forever until cancelled.
+actor CancellableContinuation {
+    /// The continuation.
+    private var continuation: CheckedContinuation<Void, Never>?
+
+    /// This method suspends forever until it is cancelled.
+    func run() async throws {
```
fabianfett, line 21: Why does this throw?

Sources/ServiceLifecycle/ServiceRunner.swift
```diff
@@ -0,0 +1,239 @@
...
+                    }
+                }
+            }
+
+            var finishedServiceCount = 0
+
+            // We are going to wait for any of the services to finish or
+            // the signal sequence to throw an error.
+            while !group.isEmpty {
+                let result = try await group.next()
```
fabianfett, line 132: I think we need to use `nextResult()!` here. As a throw would leave this scope!

Sources/ServiceLifecycle/ServiceRunner.swift
```diff
@@ -0,0 +1,239 @@
...
+
+                case .signalSequenceFinished:
+                    // This can happen when we are either cancelling everything or
+                    // when the user did not specify any shutdown signals. We just have to tolerate
+                    // this.
+                    continue
+
+                case nil:
+                    // This is a bit weird but I guess it can happen
+                    return
```
fabianfett, line 230: I don't think this can happen. we checked if the group is empty in the while loop.

Sources/ServiceLifecycle/ServiceRunner.swift
```diff
@@ -0,0 +1,239 @@
...
+                    for service in self.services.reversed() {
+                        self.logger.debug(
+                            "Shutting down service",
+                            metadata: [
+                                self.configuration.loggingConfiguration.serviceKey: "\(service)",
+                            ]
+                        )
+
+                        do {
+                            try await service.shutdownGracefully()
```
fabianfett, line 205: With the default implementation that we provide for Service this would hang forever:

```swift
extension Service {
     public var isLongRunning: Bool {
         // By default all services are treated as long running services.
         true
     }

     public func run() async throws {
         // We are just going to suspend here until we get cancelled.
         try await CancellableContinuation().run()
     }

     public func shutdownGracefully() async throws {
         // no-op
     }
}
```

Thinking more about this, I have the feeling that we work around a limitation in TaskGroup with the `shutdownGracefully`. The problem is that taskGroup does not allow us to cancel child tasks individually. However I think we should not leak this deficiency in task group into our API with `shutdownGracefully`. Instead for now we should use unstructured Tasks inside the ServiceRunner and add a big feature request for the Swift team.

Sources/ServiceLifecycle/ServiceRunner.swift
```diff
@@ -0,0 +1,239 @@
...
+
+            case .running:
+                self.state = .finished
+            }
+
+        case .running:
+            throw ServiceRunnerError.alreadyRunning()
+
+        case .finished:
+            throw ServiceRunnerError.alreadyFinished()
```
fabianfett, line 77: Adding the file and line number here is nice for us as library authors, but for users of the lib it would be _much_ nicer to get the file and line number where the adopter has invoked the run method. Because in those cases adopters have messed up their run methods and it would be extremely cool to tell them directly where.

Sources/ServiceLifecycle/ServiceRunner.swift
```diff
@@ -0,0 +1,239 @@
...
+            // First we have to register our signals.
+            let unixSignals = await UnixSignalsSequence(trapping: self.configuration.gracefulShutdownSignals)
+
+            group.addTask {
+                for await signal in unixSignals {
+                    return .signalCaught(signal)
+                }
+
+                return .signalSequenceFinished
+            }
```
fabianfett, line 107: Would it make more sense to check if `unixSignals` is not empty before we start that task? In this case we could remove the `signalSequenceFinished`. and just fatal after the for loop.

fabianfett (commented):
I'm extremely excited about this and a big +1. Thanks for driving this @FranzBusch! However there are some details that we should discuss:

Sources/ServiceLifecycle/Service.swift
```diff
@@ -0,0 +1,51 @@
...
+    ///
+    /// Concrete implementation should execute their long running work in this method such as:
+    /// - Handling incoming connections and requests
+    /// - Background refreshes
+    func run() async throws
+
+    /// This method is called when the ``ServiceRunner`` receives a graceful shutdown signal.
+    ///
+    /// Concrete implementations can implement this if they support graceful shutdown.
+    func shutdownGracefully() async throws
```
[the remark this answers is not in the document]
FranzBusch: Task cancellation and `shutdownGracefully` serve two entirely different purposes. Task cancellation is a _forceful_ action where every child task end their work as soon as possible, e.g. most `AsyncSequence`s are returning `nil` or throw `CancellationError` when the consuming task is cancelled.
The purpose behind `shutdownGracefully` is to allow (mostly) servers to implement quiescing logic. While we could just cancel the task when we receive a graceful shutdown signal this would leave the users almost no way to implement any custom quiescing logic since their task just got cancelled.

Imagine the following scenario. We have a gRPC server that does bidirectional streaming of `Foo` objects. When we want to gracefully shutdown the the server our business logic wants to send out the latest state of `Foo` and a message indicating that the server is bouncing. If we would use task cancellation here then the task that is currently handling the bidirectional stream will just get cancelled. That's why we need a separate mechanism to inform the users of such an event. 

Tl;dr: task cancellation != graceful shutdown

Side note: It would be create if we could get something like structured event handling into the stdlib that allows for the same pattern as `withTaskCancellation` but for arbitrary events.

Sources/ServiceLifecycle/CancellableContinuation.swift
FranzBusch: It is very normal for `run()` method implementations to throw. What NIO based servers often do in run is call `bind` which can throw. There are other things that users might want to do in here that can throw such as iterating a throwing  `AsyncSequence`.

Sources/ServiceLifecycle/ServiceRunner.swift
FranzBusch: This is a bit annoying. No child task is actually throwing here. We should probably just use `try!`

Sources/ServiceLifecycle/ServiceRunner.swift
FranzBusch: I guess we could just forward `line` and `file` here.

Sources/ServiceLifecycle/ServiceRunner.swift
FranzBusch: I am a bit hesitant to make that change. The `signal` sequence might finish if for example the signal handler get overridden.

Sources/ServiceLifecycle/ServiceRunner.swift
FranzBusch: This is not what we are trying to solve. See the comment above w.r.t. to why we need `shutdownGracefully`

Sources/ServiceLifecycle/ServiceRunner.swift
FranzBusch: I agree it shouldn't happen :D

Sources/ServiceLifecycle/Service.swift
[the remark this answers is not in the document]
FranzBusch: Though we can discuss if we want to change the pattern to be `withShutdownGracefulHandler` instead of an explicit method. We can make that work outside the stdlib probably.

Sources/ServiceLifecycle/CancellableContinuation.swift
fabianfett: True but this is an explicit implementation that does not throw :)

Sources/ServiceLifecycle/ServiceRunner.swift
fabianfett: Fine, but then we should not have a default implementation for `shutdownGracefully` that does nothing :)

Sources/ServiceLifecycle/CancellableContinuation.swift
FranzBusch: Ah sorry, misread where the comment is.

Sources/ServiceLifecycle/ServiceRunner.swift
FranzBusch: I agree that the default implementations are tricky. Maybe we shouldn't have any at all?

Sources/ServiceLifecycle/Service.swift
```diff
@@ -0,0 +1,51 @@
...
+//===----------------------------------------------------------------------===//
+
+/// This is the basic protocol that a service has to implement.
+public protocol Service: Sendable {
+    /// This indicates if the service is expected to be running for the entire duration.
+    /// If a long running service is returning from the ``Service/run()-1ougx`` method it is treated
+    /// as an unexpected early exit and all other services are going be cancelled.
+    ///
+    /// By default this is set to `true`.
+    var isLongRunning: Bool { get }
```
FranzBusch, line 22: I am not the biggest fan of this one. Especially how it interacts with the `run` method. I was thinking that we could introduce two protocols. One for long running services and the other for short running services. Open to suggestions here!

tomerd:
thanks for putting together @FranzBusch looks very exciting and welcome advancement. will take a deeper look in a week or so, but please continue to iterate with the team until then

tomerd:

Package.swift
```diff
@@ -1,34 +1,60 @@
...
+            targets: ["UnixSignals"]
+        ),
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-atomics.git", from: "1.0.0"),
-        .package(url: "https://github.com/apple/swift-log.git", from: "1.0.0"),
-        .package(url: "https://github.com/apple/swift-metrics.git", "1.0.0" ..< "3.0.0"),
-        .package(url: "https://github.com/swift-server/swift-backtrace.git", from: "1.1.1"),
-        .package(url: "https://github.com/apple/swift-nio.git", from: "2.0.0"), // used in tests
-        .package(url: "https://github.com/apple/swift-docc-plugin", from: "1.0.0"),
```
tomerd, line 17: emitting metric on application start and shutdown has been _very_ valuable in real world production services, the canonical example is a monitoring this to detect an a spike in application restart because of crashes

Sources/ServiceLifecycle/Service.swift
tomerd: _if_ we need to know this to drive behavior, then a protocol seems more flexible and future facing

Sources/ServiceLifecycle/Service.swift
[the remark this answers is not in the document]
fabianfett: Okay. Fine with me. Can we add this discussion GracefulShutdown vs. TaskCancellation as docu? And should users still implement Cancellation?

Sources/ServiceLifecycle/ServiceRunner.swift
fabianfett: > Maybe we shouldn't have any at all?

+1

Package.swift
fabianfett: Generally I agree. But ServiceLifecycle created the weird situation in the past where ServiceLifecycle depended on a MetricSystem that adopters wanted to setup with ServiceLifecycle. Basically the first use occurred even before the system was actually ready to create Counters. A good example for this are MetricSystems that are sending their data to a server themselves.

Sources/ServiceLifecycle/Service.swift
PeterAdams-A: Not sure I'm a massive fan of the name of this.  Also wonder if there is a need for it.

Package.swift
tomerd: I can see your point. Maybe the solution is for this library to take an abstract handler that can emit the specific metrics, or even something like a delegate so that the application / service can do the bridging

FranzBusch:
Thanks for the great feedback already. I just pushed a new change that removes the `shutdownGracefully` method and the default implementations. The new approach is very similar to how task cancellation listening works.

```swift
await withShutdownGracefulHandler {
    // Your work here
} onGracefulShutdown: {
    // Your graceful shutdown logic here
}
```

The great thing is that this automatically propagates to child tasks since it is backed by a task local. This means that server authors mostly just have to start the quiescing on their NIO based channel and any work that happens in for example a gRPC streaming request will be able to listen to graceful shutdown without any manual work! I think this works out amazingly!

I would love to get feedback on this. The other thing where I am still a bit torn is the long running vs non long running work. Using a protocol approach also feels brittle because once you conform to the `LongRunningService` protocol you cannot remove it anymore without breaking API. My current proposal would be to just remove that flag altogether and expect every `run()` method to be long running. I haven't come across a need for it myself and we can always add it in later.

Lukasa:
Nit on the name: `withGracefulShutdownHandler`.

Sources/ServiceLifecycle/ServiceRunner.swift
```diff
@@ -0,0 +1,218 @@
...
+                    }
+                }
+
+                var finishedServiceCount = 0
+
+                // We are going to wait for any of the services to finish or
+                // the signal sequence to throw an error.
+                while !group.isEmpty {
+                    // No child task is actually throwing here so the try! is safe
+                    let result = try! await group.next()
```
glbrntt, line 135: super-minor: we could use a non-throwing-`TaskGroup` and return a `Result<Void, Error>` from the group instead if we really wanted to avoid the `!`

Sources/ServiceLifecycle/ServiceRunner.swift
```diff
@@ -0,0 +1,218 @@
...
+
+                    switch result {
+                    case .serviceFinished(let service):
+                        if service.isLongRunning {
+                            // If a long running service finishes early we treat this as an unexpected
+                            // early exit and have to cancel the rest of the services.
+                            self.logger.error(
+                                "Service finished unexpectedly",
```
glbrntt, line 143: Probably worth mentioning that it's a long-running service to hint back to the API

Sources/ServiceLifecycle/ServiceRunner.swift
```diff
@@ -0,0 +1,218 @@
...
+                        // services. This allows the users to declare a hierarchy with the order they passed
+                        // the services.
+                        self.logger.info(
+                            "Signal caught. Shutting down services",
+                            metadata: [
+                                self.configuration.loggingConfiguration.signalKey: "\(unixSignal)",
+                            ]
+                        )
+
+                        await shutdownGracefullyManager.shutdownGracefully()
```
glbrntt, line 199: Do we want to allow this to run indefinitely?

Sources/ServiceLifecycle/ServiceRunnerConfiguration.swift
```diff
@@ -0,0 +1,49 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import UnixSignals
+
+/// The configuration for the ``ServiceRunner``.
+public struct ServiceRunnerConfiguration: Hashable, Sendable {
+    /// The runner's logging configuration.
+    public struct LoggingConfiguration: Hashable, Sendable {
```
glbrntt, line 20: Is this just for keys? If so we should probably change the name of this config object to include `Keys` and then we can drop the `Key` suffix from each property.

Sources/ServiceLifecycle/ServiceRunnerConfiguration.swift
```diff
@@ -0,0 +1,49 @@
...
+
+        /// Initializes a new ``ServiceRunnerConfiguration/LoggingConfiguration-swift.struct``.
+        public init() {}
+    }
+
+    /// The signals that lead to graceful shutdown.
+    public var gracefulShutdownSignals: [UnixSignal]
+
+    /// The runner's logging configuration.
+    public var loggingConfiguration: LoggingConfiguration
```
glbrntt, line 40: I think we can drop the `Configuration` suffix from the property name: it should be implicit given it's a property on a config object.

Sources/ServiceLifecycle/ServiceRunnerConfiguration.swift
```diff
@@ -0,0 +1,49 @@
...
+
+    /// Initializes a new ``ServiceRunnerConfiguration``.
+    ///
+    /// - Parameter gracefulShutdownSignals: The signals that lead to graceful shutdown.
+    public init(gracefulShutdownSignals: [UnixSignal] = []) {
```
glbrntt, line 45: Does the default make it harder for users to know about graceful shutdown? Not providing a default would make it easier to discover.

Sources/ServiceLifecycle/ServiceRunnerError.swift
```diff
@@ -0,0 +1,127 @@
...
+                line: line
+            )
+        )
+    }
+
+    /// Indicates that the service runner has already finished running.
+    public static func alreadyFinished(file: String = #fileID, line: Int = #line) -> Self {
+        Self(
+            .init(
+                errorCode: .alreadyRunning,
```
glbrntt, line 104: this should be `.alreadyFinished`

Sources/UnixSignals/UnixSignalsSequence.swift
```diff
@@ -0,0 +1,311 @@
...
+/// An unterminated `AsyncSequence` of ``UnixSignal``s.
+///
+/// This can be combined with a `TaskGroup` to signal cancellation by trapping Unix signals. As an
+/// example:
+///
+/// ```
+/// @main
+/// struct Server {
+///   static func main() async throws {
+///     await withTaskGroup(of: Void.self) { group in
```
glbrntt, line 31: Suggest making this example less service-lifcecycley

Sources/ServiceLifecycle/Service.swift
```diff
@@ -0,0 +1,30 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// This is the basic protocol that a service has to implement.
+public protocol Service: Sendable {
+    /// This indicates if the service is expected to be running for the entire duration.
+    /// If a long running service is returning from the ``Service/run()-1ougx`` method it is treated
+    /// as an unexpected early exit and all other services are going be cancelled.
```
glbrntt, line 19: "unexpected early exit" sounds a lot like `throws`... what's an example of a non-long-lived service and why do we need to make this distinction?

Sources/ServiceLifecycle/GracefulShutdown.swift
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    func removeHandler(_ handlerNumber: UInt64) {
+        self.handlers.removeAll { $0.0 == handlerNumber }
+    }
+
+    @_spi(Testing)
+    public func shutdownGracefully() {
+        guard !self.isShuttingDown else {
+            fatalError("Tried to shutdown gracefully more than once")
```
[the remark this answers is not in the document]
FranzBusch: I changed it because we need APIs that allow users to trigger this as well for testing purposes

Sources/ServiceLifecycle/Service.swift
```diff
@@ -0,0 +1,30 @@
...
+    ///
+    /// By default this is set to `true`.
```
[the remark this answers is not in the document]
FranzBusch: I removed that for now but I would like to discuss this more. I am not sure we need this property at all.

Sources/ServiceLifecycle/ServiceRunner.swift
FranzBusch: Why run `indefinitely`? The `wait` part is only there to hop to the underlying actor the handlers are all `sync`

Sources/ServiceLifecycle/ServiceRunnerConfiguration.swift
FranzBusch: I wanted to keep it open for more logging config if we ever want to add something.

Sources/ServiceLifecycle/ServiceRunner.swift
glbrntt: Ah, I missed that, I thought we were awaiting for the shutdown to complete.

Sources/ServiceLifecycle/GracefulShutdown.swift
```diff
@@ -0,0 +1,154 @@
...
+    public static var gracefulShutdownManager: GracefulShutdownManager?
+}
+
+extension Task {
+    /// Indicates that the task should shutdown gracefully.
+    ///
+    /// Calling this method on a task that doesn’t support graceful shutdown has no effect.
+    ///
+    /// - Note: This method is mostly relevant for testing graceful shutdown. In your application, the ``ServiceRunner``
+    /// should be the instance that triggers the graceful shutdown.
```
fabianfett, line 53: How do I use this in a test case?

fabianfett (commented):
Looks much nicer already! Will go in depth tomorrow.

Sources/ServiceLifecycle/Docs.docc/How to adopt ServiceLifecycle in libraries.md
```diff
@@ -0,0 +1,126 @@
...
+When running an application in a real environment it is often required to gracefully shutdown the application.
+For example the application might be running in Kubernetes and a new version of it got deployed. In this
+case, Kubernetes is going to send a `SIGTERM` signal to the application and expects it to terminate
+within a grace period. If the application isn't stopping in time then Kubernetes will send the `SIGKILL`
+signal and forcefully terminate the process.
+For this reason ``ServiceLifecycle`` introduce a new _shutdown gracefully_ concept that allows to
+terminate the work in a structured and graceful manner. This works similar to task cancellation but
+it is fully opt in and up to the business logic of the application to decide what to do.
+
+``ServiceLifecycle`` exposes one free function called ``withShutdownGracefulHandler(operation:onGracefulShutdown:)``
```
tachyonics, line 110: It would be useful to have an example here of how a service would incorporate this method.

Sources/ServiceLifecycle/ServiceRunner.swift
```diff
@@ -0,0 +1,198 @@
...
+
+                for service in self.services {
+                    self.logger.debug(
+                        "Starting service",
+                        metadata: [
+                            self.configuration.logging.serviceKey: "\(service)",
+                        ]
+                    )
+
+                    group.addTask {
```
tachyonics, line 121: Currently there isn't a way for a service to indicate the priority it needs to run at which seems important for certain classes of services - logging, metrics backends for example.

FranzBusch:
Just discussed this some more with @fabianfett. 

- With the new `withGracefulShutdownHandler` approach I missed making sure all services are shutting down in order. To fix this we should send the graceful shutdown signal down each service child task and wait for the respective run method to finish before sending it to next child task.
- The testing interface that I created is not really fitting. I will experiment with a `with` based API for testing instead
- Oftentimes we have services that have `run()` methods where they want to treat graceful shutdown similar to cancellation. These methods are often iterating an `AsyncSequence`. We should provide convenience APIs here to make this work.

adam-fowler:
> Just discussed this some more with @fabianfett.
> 
> * With the new `withGracefulShutdownHandler` approach I missed making sure all services are shutting down in order. To fix this we should send the graceful shutdown signal down each service child task and wait for the respective run method to finish before sending it to next child task.

Services don't have an initialising state, and could therefore all be initialising at the same time. As soon as one service initialisation hits a suspension point another service can start initialising. So being careful about shutdown ordering doesn't seem that important.

Sources/ServiceLifecycle/GracefulShutdown.swift
```diff
@@ -0,0 +1,154 @@
...
+/// A common use-case is to listen to graceful shutdown and use the `ServerQuiescingHelper` from `swift-nio-extras` to
+/// trigger the quiescing sequence. Furthermore, graceful shutdown will propagate to any child task that is currently executing
+///
+/// - Parameters:
+///   - operation: The actual operation.
+///   - handler: The handler which is invoked once graceful shutdown has been triggered.
+@_unsafeInheritExecutor
+public func withShutdownGracefulHandler<T>(
+    operation: () async throws -> T,
+    onGracefulShutdown handler: @Sendable @escaping () -> Void
```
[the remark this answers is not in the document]
FranzBusch: I think you should keep the state machine. It is in general just good to have that to reduce state handling error.

W.r.t. to async or not for the closure. I followed the pattern that onCancel established but we could also make it async. Open for input here.

I understand that this would result in another unstructured Task to communicate to your actor otherwise.
As a general note, the onGracefulShutdown is not running on your actor executor but on the one calling graceful shutdown. So you still need to hop on your actor before doing state modifications

FranzBusch:
> Services don't have an initialising state, and could therefore all be initialising at the same time. As soon as one service initialisation hits a suspension point another service can start initialising. So being careful about shutdown ordering doesn't seem that important.

You are correct that services don't have initialization stages because that is happening before they are passed to the runner.
However, it is super important that we call the shutdown handlers in the reverse order and make sure a service is shutdown before we go to the next one.

Simple example, you have an HTTPclient and a gRPC server. In your server you are using the http client to make requests on a RPC call. We need to make sure the client is running until the server is done. Otherwise, you might not be able to handle a gRPC request because the client is already shutdown.

Sources/ServiceLifecycle/Docs.docc/How to adopt ServiceLifecycle in libraries.md
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  private func sendKeepAlivePings() async { ... }
+}
+```
+
+
+### Returning from your `run()` method
+
+Since the `run()` method contains long running, work returning from it is seen as a failure and will
```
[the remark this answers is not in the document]
FranzBusch: I think this was mostly about the wrongly places `,`. It should have been after `work`

Sources/ServiceLifecycle/ServiceRunnerConfiguration.swift
```diff
@@ -0,0 +1,49 @@
...
+        public var signalsKey = "signals"
+        /// The logging key used for logging the service.
+        public var serviceKey = "service"
+        /// The logging key used for logging the services.
+        public var servicesKey = "services"
+        /// The logging key used for logging an error.
+        public var errorKey = "error"
+
+        /// Initializes a new ``ServiceRunnerConfiguration/LoggingConfiguration``.
+        public init() {}
```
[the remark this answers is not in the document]
FranzBusch: The default synthesised initialisers are always at `internal` level.

Sources/ServiceLifecycle/ServiceRunner.swift
FranzBusch: Interesting point. I think priority in our ecosystem is a bit of an odd thing since a lot of services are going to be backed by NIO channels & ELs. At that point, the priority is kinda lost because these threads don't use priorities.

My current thinking is to leave priorities out for the initial release. We can add them in later through either extending the `Service` protocol or making the configuration more sophisticated. WDYT?

tachyonics: We have definitely been able to improve our worst case latencies under load by considering priority. I agree with your point about priority when it gets the NIO layer but for instance our metrics backend uses an async sequence to batch and throttle the metric events before trying to send them across the network. So the original events are potentially an order of magnitude higher frequency than what is being handled by an EL.

It may come down to load testing but potentially not considering priority will block us adopting this across the board.

FranzBusch: I am curious to how you schedule your different services right now. Are you using a task group to run all of them?

The background for my question is that priority escalation should kick in here and actually make your lower priority service run at the higher priority again. Simple example

```swift
Task(priority: .medium) {
  await withTaskGroup(of: Void.self) { group in
    group.add(priority: .low) {
        await metricsService.run()
    }
    
    await group.next() // This will escalate the priority of the child task to `medium` 
  }
}
```

Sources/ServiceLifecycle/GracefulShutdown.swift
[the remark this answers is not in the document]
FranzBusch: Having thought about this again. I think `onGracefulShutdown` should be non `async`. The reason for this is that we should follow the pattern of how cancellation works. This is due to the fact on how actually triggering graceful shutdown also works. If we call `GracefulShutdownManager.shutdownGracefully()` we can't be sure that the tasks actually shutdown after that call. We can only be sure that it finished once the child tasks return.

Probably the thing we want to do here is use and unstructured task in the `onGracefulShutdown` closure and call back to the actor server which will then do the right thing. In the future, we might be able to avoid that unstructured task by having `send` semantics for actor method invocations cc @ktoso

Sources/ServiceLifecycle/ServiceRunner.swift
FranzBusch: To follow up on this. I have just checked how task priorities work on Linux and they are basically a no-op. There is no API on Linux Dispatch to implement this.

So I am surprised that you saw an improvement of latencies. Did you measure this on Linux or Darwin?

Sources/ServiceLifecycle/GracefulShutdown.swift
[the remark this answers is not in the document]
adam-fowler: So are you saying we shouldn't make `onGracefulShutdown` async because it might cause confusion and imply that once it returns it has completed shutdown?

It seems odd though that everything else about this module points the user to use structured concurrency, but as soon as we get to shutdown the recommendation is use an unstructured task.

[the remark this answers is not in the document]
FranzBusch: Fully agree with you; however, this is just a missing thing in the language IMO. We already have this problem with `onCancel` and `onGracefulShutdown` is just going to add to that list. Actors should provide a way to send them a message that is ordered correctly and doesn't await a response.

I wouldn't want to make the handler `async` to just support that case.

FranzBusch:
> One of the nice things that the original service-lifecycle had was you could create a child service-lifecycle ComponentLifecycle consisting of multiple services and register that with your main service-lifecycle. The main service-lifecycle would then manage the lifecycle of these services in the child thus creating a hierarchy of services instead of a flat array. I'm not sure how you would do that with the new implementation.

I have thought about this a bit as well and IMO we don't need any special thing here. Since we have Structured Concurrency you can just create a new `ServiceRunner` inside another service and use it. It is just important to not set the `gracefulShutdownSignal`s. Structured Concurrency is making sure the whole parent to child relationship is in place and works. It might be that we need more APIs on the runner itself to support this super nicely, but I think we should reduce the amount of APIs first and then come with concrete use-cases and expand them again.

Sources/ServiceLifecycle/Service.swift
```diff
@@ -0,0 +1,51 @@
...
+
+extension Service {
+    public var isLongRunning: Bool {
+        // By default all services are treated as long running services.
+        true
+    }
+
+    public func run() async throws {
+        // We are just going to suspend here until we get cancelled.
+        try await CancellableContinuation().run()
```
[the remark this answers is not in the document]
FranzBusch: This has been resolved by not defaulting run and removing the notion of non long running services completely.

FranzBusch:
I have updated all the code to use the new `withGracefulShutdownHandler` and provided convenience methods on `AsyncSequence` to cancel the iteration on graceful shutdown. Furthermore, I introduced a new `ServiceLifecycleTestKit` that allows testing of the graceful shutdown. Lastly, I added a new documentation article for application authors.

I would love to get another round of feedback so that we can move in the direction of merging this and tagging an alpha version to try adopting it in some libraries. I also want to follow up with a concrete NIO based example in this repository once we have landed the structured concurrency support in NIO.

cc @fabianfett @glbrntt @tomerd @adam-fowler @tachyonics @gjcairo

Sources/UnixSignals/UnixSignalsSequence.swift
```diff
@@ -17,6 +17,7 @@ import Darwin
 #elseif canImport(Glibc)
 import Glibc
 #endif
+import ConcurrencyHelpers
```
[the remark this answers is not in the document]
FranzBusch: Yes we need it to get the `LockedValueBox`

Sources/ConcurrencyHelpers/LockedValueBox.swift
```diff
@@ -11,6 +11,19 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 //===----------------------------------------------------------------------===//
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
```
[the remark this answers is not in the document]
weissi: Yes, you need to reproduce the exact NIO one and you can/should add your own above the original (NIO) one.

Sources/ServiceLifecycle/GracefulShutdown.swift
```diff
@@ -0,0 +1,116 @@
...
+/// will be executed immediately before operation gets to execute. This allows the `onGracefulShutdown` handler to set some external “shutdown” flag
+/// that the operation may be atomically checking for in order to avoid performing any actual work once the operation gets to run.
+///
+/// A common use-case is to listen to graceful shutdown and use the `ServerQuiescingHelper` from `swift-nio-extras` to
+/// trigger the quiescing sequence. Furthermore, graceful shutdown will propagate to any child task that is currently executing
+///
+/// - Parameters:
+///   - operation: The actual operation.
+///   - handler: The handler which is invoked once graceful shutdown has been triggered.
+@_unsafeInheritExecutor
```
fabianfett, line 30: Do we really want that?

FranzBusch: Some offline discussion here:

I tried to use `@_inheritActorContext` first here but it didn't work. The magic piece is to also mark the closure as `@Sendable` at the same time. This solves the warnings; however, this makes the `withGracefulShutdown` method only work inside actors. Let's look at concrete code

```swift
final actor Foo {
    var bar = 1
    func run() async {
        await withShutdownGracefulHandler {
            bar += 1
        } onGracefulShutdown: {
        }
    }
}

final class FooBar {
    var bar = 1
    func run() async {
        await withShutdownGracefulHandler {
            self.bar += 1 // Produces a warning: Capture of 'self' with non-sendable type 'FooBar' in a `@Sendable` closure
        } onGracefulShutdown: {
        }
    }
}
```

The thing that we want to convey though is that the `operation` closure of the `withShutdownGracefulHandler` is running on the same executor as the calling context. That's why I reached for `@_unsafeInheritExecutor`; however, as it turns out that thing really is unsafe and doesn't hop at all anymore.

Conclusion:
For now the best thing to do is use the `@_inheritActorContext` and `@Sendable` on the `operation` closure which forces people to use this only in actors. I am also going to create an issue on the compiler that we need to be able to spell this kind of `with` method without getting warnings.

Sources/ServiceLifecycle/ServiceRunner.swift
```diff
@@ -0,0 +1,265 @@
...
+                    }
+                }
+            }
+
+            precondition(gracefulShutdownManagers.count == self.services.count, "We did not create a graceful shutdown manager per service")
+
+            // We are going to wait for any of the services to finish or
+            // the signal sequence to throw an error.
+            while !group.isEmpty {
+                // No child task is actually throwing here so the try! is safe
```
fabianfett, line 145: code comment outdated.

Sources/ServiceLifecycle/ServiceRunner.swift
```diff
@@ -0,0 +1,265 @@
...
+
+                            return .failure(error)
+
+                        case .signalCaught, .signalSequenceFinished:
+                            fatalError("Signal sequence already returned a signal.")
+
+                        case nil:
+                            fatalError("Invalid result from group.next().")
+                        }
+                    }
```
fabianfett, line 244: maybe move this into its own method?

Sources/ServiceLifecycle/ServiceRunner.swift
```diff
@@ -0,0 +1,265 @@
...
+                case .serviceThrew(let service, _, let error):
+                    // One of the servers threw an error. We have to cancel everything else now.
+                    self.logger.error(
+                        "Service threw error. Cancelling all other services now",
+                        metadata: [
+                            self.configuration.logging.serviceKey: "\(service)",
+                            self.configuration.logging.errorKey: "\(error)",
+                        ]
+                    )
+                    group.cancelAll()
```
fabianfett, line 171: Do we always want that? Especially the cancel? We might have unimportant services that, if they can't do their job anymore, we want to gracefully shutdown, but not cancel stuff that is in flight?

Sources/ServiceLifecycle/ServiceRunner.swift
```diff
@@ -0,0 +1,265 @@
...
+
+                case .serviceThrew(let service, _, let error):
+                    // One of the servers threw an error. We have to cancel everything else now.
+                    self.logger.error(
+                        "Service threw error. Cancelling all other services now",
+                        metadata: [
+                            self.configuration.logging.serviceKey: "\(service)",
+                            self.configuration.logging.errorKey: "\(error)",
+                        ]
+                    )
```
fabianfett, line 170: how do we think about using the error log level here? In an sswg logging doc we say, that normally libs should not log higher than debug.

Sources/ServiceLifecycle/ServiceRunner.swift
```diff
@@ -0,0 +1,265 @@
...
+                    // If a long running service finishes early we treat this as an unexpected
+                    // early exit and have to cancel the rest of the services.
+                    self.logger.error(
+                        "Service finished unexpectedly. Cancelling all other services now",
+                        metadata: [
+                            self.configuration.logging.serviceKey: "\(service)",
+                        ]
+                    )
+
+                    group.cancelAll()
```
fabianfett, line 159: Do we always want that? Especially the cancel? We might have unimportant services that, if they can't do their job anymore, we want to gracefully shutdown, but not cancel stuff that is in flight?

Sources/ServiceLifecycle/ServiceRunner.swift
```diff
@@ -0,0 +1,265 @@
...
+
+                        case .serviceThrew(let service, _, let error):
+                            self.logger.error(
+                                "Service threw error during graceful shutdown. Cancelling all other services now",
+                                metadata: [
+                                    self.configuration.logging.serviceKey: "\(service)",
+                                    self.configuration.logging.errorKey: "\(error)",
+                                ]
+                            )
+                            group.cancelAll()
```
fabianfett, line 234: if one service throws on shutdown we want to escalate all following graceful shutdowns to cancel. this seems quite aggressive

Sources/ServiceLifecycle/ServiceRunnerError.swift
```diff
@@ -0,0 +1,127 @@
...
+        /// Indicates that the service runner is already running.
+        public static let alreadyRunning = Code(code: .alreadyRunning)
+        /// Indicates that the service runner has already finished running.
+        public static let alreadyFinished = Code(code: .alreadyFinished)
+        /// Indicates that a service finished unexpectedly.
+        public static let serviceFinishedUnexpectedly = Code(code: .serviceFinishedUnexpectedly)
+    }
+
+    /// Internal class that contains the actual error code.
+    private final class Backing: Hashable, Sendable, CustomStringConvertible {
```
fabianfett, line 51: I don't think `Backing` should implement `CustomStringConvertible`. I think the error itself should implement `CustomStringConvertible`.

Sources/ServiceLifecycle/ServiceRunner.swift
```diff
@@ -0,0 +1,265 @@
...
+                    "Starting service",
+                    metadata: [
+                        self.configuration.logging.serviceKey: "\(service)",
+                    ]
+                )
+
+                let gracefulShutdownManager = GracefulShutdownManager()
+                gracefulShutdownManagers.append(gracefulShutdownManager)
+
+                group.addTask {
```
fabianfett, line 128: can we add a comment here that this must remain addTask and should _not_ become `addTaskUnlessCancelled` to ensure that we close all services :)

fabianfett (commented):
Looks already very good. Couple of questions...

Sources/ServiceLifecycle/ServiceRunner.swift
FranzBusch: This is basically coming back to the discussion of short vs long running services. After discussing this with many people I haven't heard a good example of a short running service that doesn't result int cancellation. If you have a service that is unimportant e.g. a service that sends logs out to some backend. Then you probably should model that service in a way that it just buffers logs and retries the connection.

In general, I would like to go forward without supporting short running services and revisit that topic later once we have concrete cases where we know we need this. Adding support for this can be done in a non-breaking change so I don't see it as a blocker.

sliemeobn:
I love where this is going, and this is exactly the sort of crispy cleanness I was hoping to see when structured concurrency starts arriving at the server-side. Thank you!

I have a question about composition, similar to what @adam-fowler had asked:

I understand that just chaining service runners down the task-tree can be pretty much all we want, but I am missing how one would propagate "graceful shutdown" at the moment. I was playing with this code snippet:

```
struct Main {
    static func main() async throws {
        let runner = ServiceRunner.init(services: [Bar()],
                                        configuration: .init(gracefulShutdownSignals: [.sigint, .sigterm]),
                                        logger: Logger(label: "Top"))

        try await runner.run()
    }
}

struct Bar: Service {
    func run() async throws {
        try await ServiceRunner(
            services: [Foo(), Foo()], //a few sub-services
            configuration: .init(gracefulShutdownSignals: []),
            logger: Logger(label: "Bar")
        ).run()
    }
}

struct Foo: Service {
    func run() async throws {
        let isShutdown = LockedValueBox(false)
        try await withShutdownGracefulHandler {
            while(!isShutdown.withLockedValue { $0 }) {
                try await Task.sleep(for: .seconds(1))
            }
        } onGracefulShutdown: {
            isShutdown.withLockedValue { $0 = true }
        }
    }
}

```

As it stands, Foo never gets the gracefulShutdown. Also, there is no (good) way for Bar to trigger an "internal" graceful shutdown (the only way in there is through unix signals ATM) - so even wrapping the inner runner in a `withShutdownGracefulHandler` would not help much.

Also, I would love to see the option to set up an "enlist if present" gracefulShutdownHandler. The idea is that libraries could provide components that automatically participate in graceful shutdown if there (ie: running in a service runner) - but not freak out if they run in other contexts, since there is an explicit method as well to trigger shutdown.

So, here are my thoughts/questions:
+ Shouldn't the `ServiceRunner` have an explicit "shutdownGracefully" method? Things would get a lot more pluggable it seems to me.
+ There should be a simple way to have an inner ServiceRunner participate in a parent's GracefulShutdownHandler. (eg: either unix signal, or "inherit", or "inheritIfPresent" to get optional sub-component-pluggability)
+ Are we sure it shouldn't be `withGracefulShutdownHandler`?

FranzBusch:
Thanks @sliemeobn for the feedback. Very much appreciate!

> I understand that just chaining service runners down the task-tree can be pretty much all we want, but I am missing how one would propagate "graceful shutdown" at the moment. I was playing with this code snippet:

You are right that this should work. The missing piece here is that the `ServiceRunner` itself needs to implement graceful shutdown handling. Let me add this so the composition works.

> Also, I would love to see the option to set up an "enlist if present" gracefulShutdownHandler. The idea is that libraries could provide components that automatically participate in graceful shutdown if there (ie: running in a service runner) - but not freak out if they run in other contexts, since there is an explicit method as well to trigger shutdown.

That is an interesting point, that I haven't thought about yet. However, I am also unsure how often that is going to be the case. I would love to gather some usage here with an alpha tag first. If we see that we need a change in behaviour we can do one of the two

1. Make `withGracefulShutdownHandler` not `fatalError` when in a context that doesn't support this
2. Introduce a `withPotentialGracefulShutdownHandler` or add an argument to the method.

> Are we sure it shouldn't be withGracefulShutdownHandler?

You are right. We should rename it. Will do this in one of the next commits

Sources/ServiceLifecycle/ServiceRunner.swift
FranzBusch: I changed the logging to `debug` for everything now. We could make this configurable in the future. Any error that happens is thrown from the `run()` method anyhow so users could catch and log it.

weissi: Right, I think that's more a general rule. If the error is returned _anyway_, don't log it (in non-debug levels) because the user can if they choose too.

It's a little more difficult with things that are recovered in the bowels of libraries and are not returned to the user, usually when things get retried. Think some HTTP client which fails to open a TCP connection on the first try, succeeding on the second. The user will not see an error because it could perform its job after all.
But what if, say, 50% of all TCP connections fail to establish then you likely want to know about it (despite the HTTP client being able to handle it internally by retrying). Personally, I think that can usually be done by sampling requests and having each 10,000th request or so log everything at `debug` but that requires the `Logger` to be passed down to everything or having [this](https://github.com/apple/swift-log/issues/261) work.

FranzBusch: Yeah it is a fair point and I think a discussion orthogonal to service-lifecycle. I followed the logging guidelines here now.

From my recent work on other libraries, I personally find logging always at debug/trace level in libraries to be limiting because as you said. There are sometimes cases in libraries where you hit unexpected behaviours but handle them gracefully. Logging those at debug level means that they basically never surface unless they cause business logic faults. The library could provide an error delegate for something like this but IMO it would be more straight forward to log those at a higher level so they surface and be reported back as bugs.

Though let's move that discussion over to logging :D

weissi: I'd say the user really should be in control. If you _do_ want to see the logs from a particular library, you can pass a logger with a `.debug` log level to the library. Or you can select that library using the `source` field or so.

Regardless, "basically never surface[s]" is only true if you don't sample the requests at all. Sampling over time is really powerful because say you have a failure that only happens in 1% of the cases, then you'd expect to see it once after 100 requests. If you sampled every 10,000th request, you'd expect to see it in your logs after 1M requests. For services that have significant load, this would still mean that you'd see it frequently enough. If that frequency is too low for you, maybe sample every 1,000th request. In general: Sample as high as you can without violating your latency SLOs and breaking the bank with emitting too many logs :). If you have a service that handles only like 100 requests per second, you might get a way with just logging everything at debug, depending on your latency & money (for log storage/retention) budget.

Sources/ServiceLifecycle/ServiceRunner.swift
tachyonics: I was just thinking about this as well. For example if a background credentials rotation that fails, it is likely we have some time (we rotate credentials 5 minutes before they are due to expire), before its going to become an issue. Likely what we would like to do is stop responding successfully to health checks so the instance stops receiving new requests and then drops out of the load balancer but survives long enough to successfully complete anything that is in flight.

FranzBusch: Thanks for bringing up a concrete example!

I don't think this requires any extra work in this library to support. Your service that is doing the credential refresh should expose its state somehow. Your health checking code will then check that state on every health check request and fail it when the scenario above happens.
Furthermore, the service `run()` can itself handle this by `Task.sleep`ing after the crendetial refresh failed for an appropriate amount of time. This is in my opinion very service logic specific since you need to see how long your credentials are still valid when the fresh failed and sleep for an appropriate amount of time.

In general, this goes to show that these "short"/"not essential" running services can be modelled with the current APIs. While they might require some extra work in the `run()` method they are possible to implement. This allows us to defer implementing something in this library and wait how adoption turns out.

tachyonics: This seems reasonable. It is essentially saying the correct way to shutdown an running application is for a signal to be thrown and any service that decides it is not in a good state should keep running but have some alternate mechanism to cause that signal. It certainly seems the correct behaviour for not-immediately-essential services in our use case; they have to go through the health check mechanism to protect the requests being processed by the critical service. I don't have any counter examples at the moment.

The work potentially is - completely outside of this - to define a common health check mechanism that makes this sort of signalling easy, if we consider that useful. That would probably fall more on the middleware side of things.

adam-fowler:
Would it be possible to add a public method to gracefully shutdown the service runner? 

I've been looking at running services in tests as follows
```swift
try await withThrowingTaskGroup(of: Void.self) { group in
    let serviceRunner = ServiceRunner(
        services: [myService],
        configuration: .init(gracefulShutdownSignals: [.sigterm, .sigint]),
        logger: Self.logger(function)
    )
    group.addTask {
        try await serviceRunner.run()
    }
    // wait for service to be ready
    await barrier.wait()

    group.addTask {
        try await testServer()
    }
    var iterator = group.makeAsyncIterator()
    try await iterator.next()
    group.cancelAll()
}
```

If I don't run the service inside the ServiceRunner I get a warning about there being no graceful shutdown handler, if I run the service inside a `ServiceRunner` as above then the cancel at the end will give the warning "Service finished unexpectedly". I would like to run tests without seeing any untoward messages that may confuse.

Sources/ServiceLifecycle/ServiceRunner.swift
FranzBusch: Health checking is definitely something interesting to consider (though agree outside of this). In my opinion, a good pattern is for each concrete `Service` to expose their external state e.g. `running` or `shutdown` through a property or an async sequence. This enables to implement a health check which listens to all the states and responds correctly.
In most orchestration systems repeated health check failures lead to `SIGTERM` or `SIGKILL` so this is going to end up shutting down the application.

ktoso:
I reviewed the code with @FranzBusch offline and looks good overall.

I agree that graceful shutdown must be possible to trigger manually; there's many reasons to do so other than externally triggered, a node leaving a cluster etc.

Sources/ServiceLifecycle/ServiceRunnerError.swift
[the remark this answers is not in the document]
FranzBusch: It already does that. It conforms to `Hashable` which inherits from `Equatable`

README.md
```diff
@@ -17,293 +17,18 @@ If you have a server-side Swift application or a cross-platform (e.g. Linux, mac
...
-```swift
-let foo = ...
-lifecycle.registerShutdown(
-    label: "foo",
-    .eventLoopFuture(foo.shutdown)
-)
-```
-
--------
+You can find the documentation on how to use ServiceLifecycle over [here](https://swiftpackageindex.com/swift-server/swift-service-lifecycle/main/documentation/lifecycle).
```
tomerd, line 31: there is allot of value in providing key examples right on the readme. i suggest we do that, especially since  the full documentation is a hop away

Sources/ServiceLifecycle/AsyncCancelOnGracefulShutdownSequence.swift
```diff
@@ -0,0 +1,161 @@
...
+    enum ElementOrEnd: Sendable {
+        case element(Base.Element)
+        case end
+    }
+
+    @usableFromInline
+    typealias Element = ElementOrEnd
+
+    @usableFromInline
+    let _base: Base
```
tomerd, line 122: many internal variables underscored. why? testing only?

Sources/ServiceLifecycle/Docs.docc/How to adopt ServiceLifecycle in applications.md
```diff
@@ -0,0 +1,200 @@
...
+all services in an application.
+
+## Adopting the ServiceRunner in your application
+
+This article is focusing on how the ``ServiceRunner`` works and how you can adopt it in your application.
+If you are interested in how to properly implement a service, go check out the article:  <doc:How-to-adopt-ServiceLifecycle-in-libraries>.
+
+### How is the ServiceRunner working?
+
+The ``ServiceRunner`` is just a slightly complicated task group under the hood that runs each service
```
tomerd, line 29: maybe we should call it ServiceGroup?

Sources/ServiceLifecycle/Docs.docc/How to adopt ServiceLifecycle in libraries.md
```diff
@@ -0,0 +1,153 @@
...
+
+An example implementation of a `TCPEchoServer` on a high level that supports graceful shutdown
+might look like this.
+
+```swift
+public actor TCPEchoClient: Service {
+  public init() { }
+
+  public func run() async throws {
+    await withGracefulShutdownHandler {
```
tomerd, line 124: how is graceful shutdown ordering determined in this model?

Sources/ServiceLifecycle/GracefulShutdown.swift
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    // We have to keep track of our handler here to remove it once the operation is finished.
+    let handlerID = await gracefulShutdownManager.registerHandler(handler)
+
+    let result = try await operation()
+
+    // Great the operation is finished. If we have a number we need to remove the handler.
+    if let handlerID {
+        await gracefulShutdownManager.removeHandler(handlerID)
```
tomerd, line 46: okay if this does not happen when running into an error?

Sources/ServiceLifecycle/Service.swift
```diff
@@ -0,0 +1,26 @@
...
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftServiceLifecycle project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// This is the basic protocol that a service has to implement.
+public protocol Service: Sendable {
```
tomerd, line 16: do we see this protocol grow beyond the one method?

Sources/ServiceLifecycle/ServiceRunner.swift
```diff
@@ -0,0 +1,383 @@
...
+            var potentialError: Error?
+            do {
+                try await self._run(gracefulShutdownStream: gracefulShutdownStream)
+            } catch {
+                potentialError = error
+            }
+
+            switch self.state {
+            case .initial, .finished:
+                fatalError("ServiceRunner is in an invalid state \(self.state)")
```
tomerd, line 84: unless there is a memory corruption / security risk, i am not a fan of fatalError

Sources/ServiceLifecycle/ServiceRunner.swift
```diff
@@ -0,0 +1,383 @@
...
+                            try await service.run()
+                            return .serviceFinished(service: service, index: index)
+                        } catch {
+                            return .serviceThrew(service: service, index: index, error: error)
+                        }
+                    }
+                }
+            }
+
+            precondition(gracefulShutdownManagers.count == self.services.count, "We did not create a graceful shutdown manager per service")
```
tomerd, line 212: assert better choice?

Sources/ServiceLifecycle/ServiceRunner.swift
```diff
@@ -0,0 +1,383 @@
...
+
+            // We are going to wait for any of the services to finish or
+            // the signal sequence to throw an error.
+            while !group.isEmpty {
+                let result: ChildTaskResult? = await group.next()
+
+                switch result {
+                case .serviceFinished(let service, _):
```
tomerd, line 220: this may be out of scope, but there could be type of bootstrap tasks that are long running enough so that we would want to cancel them on signal but will finish "early" or iow be short lived

one example i can think of are database migrations, it may take a few minutes and we want to be able to cancel them in an orderly manner, but they have a well defined logical end that is not the entire life of the application. further, we want them to run first, before any other tasks / services run

Sources/ServiceLifecycle/ServiceRunner.swift
```diff
@@ -0,0 +1,383 @@
...
+                    group.cancelAll()
+                    throw ServiceRunnerError.serviceFinishedUnexpectedly()
+                }
+
+            case .serviceThrew(let service, _, let error):
+                self.logger.debug(
+                    "Service threw error during graceful shutdown. Cancelling all other services now",
+                    metadata: [
+                        self.configuration.logging.serviceKey: "\(service)",
+                        self.configuration.logging.errorKey: "\(error)",
```
tomerd, line 350: general nit: not all Swift errors interpolate well, this is interesting to improve

tomerd: another general nit is that while i like the use of structured logging, need to make sure it actually outputs a useful logging statement in the default case of a "dumb" logging backend

Sources/ServiceLifecycle/ServiceRunnerConfiguration.swift
tomerd: maybe put the keys under a Keys "container" then

tomerd (approved):
only skimmed over this but the approach looks right. left a few non blocking comments, and excited to see what folks think of it when they try to integrate it with real frameworks and applications

sliemeobn:
(not sure if this is the right place for "feature requests", so I am happy to move this to a forum post or similar)

I stubbed my toes quite a bit while trying to cleanly implement the following use case, which I believe might be quite universal:

+ a service that processes message off a message queue (rabbitmq) sequentially (basic async sequence for message handling)
+ since it is "gateway server" service, it's purpose is to forward message to an external system, in this case a basic HTTP-based custom protocol
+ for robustness, on any connection error/non-200 response deliveries are retried indefinitely (with a basic timeout interval/back-off strategy)

So, having shutdown handling in the "normal" case is quite clean: just tell the message broker to "unsubscribe" from the queue -> async sequence drains and finishes -> all is good

For shutting down while in retry I'd need two things:
+ a "Task.sleep" that cancels itself on shutdown (or better a "generic block" that cancels itself on shutdown)
+ a way to know "is this task requested to shut down"

Both are are harder to do than it seems (at least for me ; ) and require a lot of ceremony at the moment.

So, I am suggesting to add something along the lines of: 
- Task.isGracefulShutdownRequested
- withCancellationOnGracefulShutdown { try await Task.sleep() }

About the second one: an awaitable signal would simplify this a lot too (eg: can be used in a task group). Much like the AsyncGracefulShutdownSequence - maybe something awaitable like this could be made public?

FranzBusch:
Thanks for all the feedback! I will circle around to responding to this next week!

Sources/ServiceLifecycle/ServiceRunner.swift
fabianfett: but this is a clear programer error isn't it?

Sources/ServiceLifecycle/AsyncCancelOnGracefulShutdownSequence.swift
FranzBusch: This is due to the usage of `@inlinable` which forces us to make these properties `internal` otherwise they would be `private`. We tend to underscore prefix these properties normally to indicate that they are not really `internal`.

Sources/ServiceLifecycle/GracefulShutdown.swift
FranzBusch: Yeah I caught this as well and fixed it up.

Sources/ServiceLifecycle/ServiceRunner.swift
FranzBusch: Exactly what @fabianfett said here. This is a programmer error on our side and we are using `fatalError`s for these quite consistently across NIO. In this particular instance if we are in any other state than `running` we made a mistake in our implementation.

Sources/ServiceLifecycle/ServiceRunner.swift
FranzBusch: A precondition here is better IMO. If we this condition isn't `true` we made a mistake in our implementation. We could also remove this precondition but it just to make sure everything is setup correctly before we proceed.

Sources/ServiceLifecycle/ServiceRunner.swift
FranzBusch: I also still see some use for short running tasks however I couldn't find a concrete example yet. The migration example is IMO not something one would model with a service because arguably the migrations should be run before any of the other services are starting. That's why I would say migrations are run before starting the service runner e.g.

```
try await migrator.run()

let serviceFoo = Foo()
let serviceBar = Bar()

try await serviceRunner([serviceFoo, serviceBar])
```

The current design leaves space for short running service in the future by letting the user pass a service configuration where they can state if a service is long or short running. This is purely additive and I would like to add that once we see a concrete need.

Sources/ServiceLifecycle/ServiceRunner.swift
FranzBusch: I agree that errors sometimes don't interpolate well but I don't think we can fix that here.

I also agree on the structured logging statement though I think this does output a useful logging statement. I rather prefer doing proper structured logging here than interpolating them into a string.

Package.swift
FranzBusch: Following up on this, I would like to get the PR merged without metrics and then add it on top since the PR is getting quite big.

Sources/ServiceLifecycle/Docs.docc/How to adopt ServiceLifecycle in applications.md
FranzBusch: I like the name `ServiceGroup` and renamed everything. Thanks!

FranzBusch:
Got around to address all the outstanding comments. Thanks again for the amazing feedback so far!

> For shutting down while in retry I'd need two things:
> 
> * a "Task.sleep" that cancels itself on shutdown (or better a "generic block" that cancels itself on shutdown)
> * a way to know "is this task requested to shut down"
> 
> Both are are harder to do than it seems (at least for me ; ) and require a lot of ceremony at the moment.
> 
> So, I am suggesting to add something along the lines of:
> 
> * `Task.isGracefulShutdownRequested`
> * `withCancellationOnGracefulShutdown { try await Task.sleep() }`


@sliemeobn I added two new APIs in the latest commit `cancelOnGracefulShutdown` and `Task.isShuttingDownGracefully` this should make it easier to implement your use-case and we saw in other projects that they are in general helpful. I also changed the default behaviours to tolerate being run outside of a `ServiceGroup` and it will just run the operation closures. I think this is the right thing because graceful shutdown is something totally optional and it might never happen.

@ktoso @tomerd @glbrntt @fabianfett I would like to move forward with merging this PR soonish and tagging a new 2.0.0-alpha for this. The current releases are 1.x.x-alpha and I think using a new SemVer major is good here to imply the significant changes of API. There are three open things:

### Support of short running services

This has come up multiple times during the review but so far we haven't found a compelling use-case for this that isn't handled with structured concurrency itself. I would like to defer solving this until we get more real world experience and concrete needs emerge. We have room to add this to the API by simply extending the `ServiceGroupConfiguration`.

### Metrics
@tomerd You brought up built-in metrics for startup and shutdown. In general, I like the idea but would like to defer this to keep the PR smaller.

### Failing checks
This package is only supporting 5..6+ so the checks for previous Swift versions are failing. Once we merge this PR, we should remove those checks.

FranzBusch:
Going ahead with merging this PR and solving the outstanding questions in follow-up PRs

PR #142, opened 2023-05-30
Rename `shutdownGracefully` to `triggerGracefulShutdown` on the `ServiceGroup`

FranzBusch:
# Motivation
The current naming of `shutdownGracefully()` on the `ServiceGroup` caused some confusion since callers had to `await` the call due the `ServiceGroup` being an `actor`. This lead users to believe that the method waited until the graceful shutdown was completed; however, the method only triggered the shutdown and exited.

# Modification
This PR renames `ServiceGroup/shutdownGracefully()` to `ServiceGroup/triggerGracefulShtudown()`.

# Result
Less confusion around the behaviour of triggering a graceful shutdown on the `ServiceGroup`.


fabianfett (approved):
Love it!

PR #147, opened 2023-08-12
Provide a `ServiceGroupConfiguration` initializer without `gracefulSh…

FranzBusch:
…utdownSignals` signals

# Motivation
When using the `ServiceGroup` in non-top-level setups such as inside a `Service.run()` method then it is confusing if you are forced to pass `gracefulShutdownSignals`.

# Modification
This PR adds a new initializer to `ServiceGroupConfiguration` which takes no arguments.

# Result
Less confusion when using the service group inside a non-top-level context.


FranzBusch:
Closing this in favour of: https://github.com/swift-server/swift-service-lifecycle/pull/148

PR #148, opened 2023-08-14
Add support for `return/ThrowBehavior` of services

FranzBusch:
# Motivation

Currently the `ServiceGroup` cancels the whole group when a service returns or throws. This is fine for most server applications but it doesn't work for CLI tools since they mostly want to use the group to orchestrate the services as long as the user command is handled and then return cleanly from the group. Furthermore, even on server setups one might want to customize the behavior of what happens when a service returns/throws e.g. one might want to shutdown the group after their HTTP server throws so that a telemetry service flushes out all the remaining data.

# Modification

This PR does a few things:
1. It creates a new `ServiceConfiguration` and `TerminationBehavior` in the `ServiceGroupConfiguration`. Those can be used to declare the services that are run and what happens when the service returns or throws
2. Adds a new `cancellationSignals` to to the `ServiceGroupConfiguration` which allows cancellation to trigger based on a signal
3. Makes sure that any given service is only retained as long as necessary i.e. when a service returns the group is not retaining it anymore. This allows freeing of resources as early as possible
4. Breaking: Removes the `Hashable` conformance on the configuration structs. This was wrong in the first place and something we should just avoid doing in general.

# Result
We now give the user even more control about the group's behaviors.

### DIFF
--- a/Sources/ServiceLifecycle/ServiceGroup.swift
+++ b/Sources/ServiceLifecycle/ServiceGroup.swift
@@ -262,7 +262,7 @@ public actor ServiceGroup: Sendable {
                         continue
                     }
 
-                    switch service.returnBehavior.behavior {
+                    switch service.successfulTerminationBehavior.behavior {
                     case .cancelGroup:
                         self.logger.error(
                             "Service finished unexpectedly. Cancelling group.",
@@ -275,7 +275,7 @@ public actor ServiceGroup: Sendable {
 
                     case .gracefullyShutdownGroup:
                         self.logger.error(
-                            "Service finished unexpectedly. Shutting down group.",
+                            "Service finished unexpectedly. Gracefully shutting down group.",
                             metadata: [
                                 self.loggingConfiguration.keys.serviceKey: "\(service.service)",
                             ]
@@ -310,7 +310,7 @@ public actor ServiceGroup: Sendable {
                     }
 
                 case .serviceThrew(let service, let index, let error):
-                    switch service.throwBehavior.behavior {
+                    switch service.failureTerminationBehavior.behavior {
                     case .cancelGroup:
                         self.logger.error(
                             "Service threw error. Cancelling group.",
@@ -363,7 +363,7 @@ public actor ServiceGroup: Sendable {
                     }
 
                 case .signalCaught(let unixSignal):
-                    if self.gracefulShutdownSignals.contains(where: { $0 == unixSignal }) {
+                    if self.gracefulShutdownSignals.contains(unixSignal) {
                         // Let's initiate graceful shutdown.
                         self.logger.debug(
                             "Signal caught. Shutting down the group.",
@@ -487,7 +487,7 @@ public actor ServiceGroup: Sendable {
                 }
 
             case .serviceThrew(let service, _, let error):
-                switch service.throwBehavior.behavior {
+                switch service.failureTerminationBehavior.behavior {
                 case .cancelGroup:
                     self.logger.error(
                         "Service threw error during graceful shutdown. Cancelling group.",
@@ -512,7 +512,7 @@ public actor ServiceGroup: Sendable {
                 }
 
             case .signalCaught(let signal):
-                if self.cancellationSignals.contains(where: { $0 == signal }) {
+                if self.cancellationSignals.contains(signal) {
                     // We got signalled cancellation after graceful shutdown
                     self.logger.debug(
                         "Signal caught. Cancelling the group.",
--- a/Sources/ServiceLifecycle/ServiceGroup.swift
+++ b/Sources/ServiceLifecycle/ServiceGroup.swift
@@ -20,7 +20,7 @@ public actor ServiceGroup: Sendable {
     /// The internal state of the ``ServiceGroup``.
     private enum State {
         /// The initial state of the group.
-        case initial
+        case initial(services: [ServiceGroupConfiguration.ServiceConfiguration])
         /// The state once ``ServiceGroup/run()`` has been called.
         case running(
             gracefulShutdownStreamContinuation: AsyncStream<Void>.Continuation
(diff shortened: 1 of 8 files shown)


Sources/ServiceLifecycle/ServiceGroup.swift
```diff
@@ -126,32 +147,46 @@ public actor ServiceGroup: Sendable {
...
             ]
         )
 
         // Using a result here since we want a task group that has non-throwing child tasks
         // but the body itself is throwing
         let result = await withTaskGroup(of: ChildTaskResult.self, returning: Result<Void, Error>.self) { group in
             // First we have to register our signals.
-            let unixSignals = await UnixSignalsSequence(trapping: self.configuration.gracefulShutdownSignals)
+            let gracefulShutdownSignals = await UnixSignalsSequence(trapping: self.gracefulShutdownSignals)
+            let cancellationSignals = await UnixSignalsSequence(trapping: self.cancellationSignals)
```
glbrntt, line 176: Might be simpler to just create one sequence because both tasks for these are exactly the same...

Sources/ServiceLifecycle/ServiceGroup.swift
```diff
@@ -200,73 +235,174 @@ public actor ServiceGroup: Sendable {
...
-                            self.configuration.logging.keys.serviceKey: "\(service)",
-                        ]
-                    )
-
-                    group.cancelAll()
-                    return .failure(ServiceGroupError.serviceFinishedUnexpectedly())
+                case .serviceFinished(let service, let index):
+                    guard !group.isCancelled else {
+                        // The group is cancelled and we expect all services to finish
+                        continue
```
glbrntt, line 262:
```suggestion
                    if group.isCancelled {
                        // The group is cancelled and we expect all services to finish
                        continue
```

Sources/ServiceLifecycle/ServiceGroup.swift
```diff
@@ -200,73 +235,174 @@ public actor ServiceGroup: Sendable {
...
+                                "All services finished."
+                            )
+
+                            group.cancelAll()
+                            return .success(())
+                        }
+                    }
+
+                case .signalCaught(let unixSignal):
+                    if self.gracefulShutdownSignals.contains(where: { $0 == unixSignal }) {
```
glbrntt, line 368: Isn't it possible to just check `self.gracefulShutdownSignals.contains(unixSignal)`?

Sources/ServiceLifecycle/ServiceGroupConfiguration.swift
```diff
@@ -41,17 +44,142 @@ public struct ServiceGroupConfiguration: Hashable, Sendable {
...
+                }
+            }
+        }
+
+        /// The service.
+        public var service: any Service
+        /// The behavior when the service returns from its `run()` method.
+        public var returnBehaviour: TerminationBehavior
+        /// The behavior when the service throws from its `run()` method.
+        public var throwBehaviour: TerminationBehavior
```
glbrntt, line 78: You need to use the US spelling for the properties here.

glbrntt: The naming here doesn't quite sit right with me; it's very strongly tied to language keywords rather than the behaviour of the service runner.

Not sure exactly what to suggest but the idea of a 'clean exit' vs. 'unclean exit' might be worth exploring.

Sources/ServiceLifecycle/ServiceGroupConfiguration.swift
```diff
@@ -41,17 +44,142 @@ public struct ServiceGroupConfiguration: Hashable, Sendable {
...
+        self.cancellationSignals = cancellationSignals
+    }
+
+
+    @available(*, deprecated)
     public init(gracefulShutdownSignals: [UnixSignal]) {
+        self.services = []
         self.gracefulShutdownSignals = gracefulShutdownSignals
-        self.logging = .init()
+        self.logger = Logger(label: "Dummy logger") // This logger will never be used
```
glbrntt, line 183: It _could_ be used I think? Would prefer setting the label to something obviously originating from here: `org.swift.service-lifecycle` or something.

Sources/ServiceLifecycle/ServiceGroup.swift
FranzBusch: I tried that first but it makes things a bit more complicated because right now I return from the child task on the first signal; however, we have to keep listening to cancellation signals once we get a graceful shutdown signal.

Sources/ServiceLifecycle/ServiceGroupConfiguration.swift
FranzBusch: Adopted the US spelling. I am really struggling to come up with something better.

Tests/ServiceLifecycleTests/ServiceGroupTests.swift
```diff
@@ -183,11 +193,154 @@ final class ServiceGroupTests: XCTestCase {
...
+
+            var eventIterator3 = service3.events.makeAsyncIterator()
+            await XCTAsyncAssertEqual(await eventIterator3.next(), .run)
+
+            await service2.resumeRunContinuation(with: .success(()))
+
+            // The last service should receive the shutdown signal first
+            await XCTAsyncAssertEqual(await eventIterator3.next(), .shutdownGracefully)
+
+            // Waiting to see that all two are still running
```
gjcairo, line 256:
```suggestion
            // Waiting to see that the remaining two are still running
```

Tests/ServiceLifecycleTests/ServiceGroupTests.swift
```diff
@@ -183,11 +193,154 @@ final class ServiceGroupTests: XCTestCase {
...
+
+            var eventIterator2 = service2.events.makeAsyncIterator()
+            await XCTAsyncAssertEqual(await eventIterator2.next(), .run)
+
+            var eventIterator3 = service3.events.makeAsyncIterator()
+            await XCTAsyncAssertEqual(await eventIterator3.next(), .run)
+
+            await service3.resumeRunContinuation(with: .success(()))
+
+            // The last service should receive the shutdown signal first
```
gjcairo, line 309:
```suggestion
            // The second service should receive the shutdown signal first
```

Tests/ServiceLifecycleTests/ServiceGroupTests.swift
```diff
@@ -183,11 +193,154 @@ final class ServiceGroupTests: XCTestCase {
...
+
+            // Let's exit from the second service
+            await service2.resumeRunContinuation(with: .success(()))
+
+            // Waiting to see that the remaining is still running
+            service1.sendPing()
+            await XCTAsyncAssertEqual(await eventIterator1.next(), .runPing)
+
+            // The first service should now receive the signal
+            await XCTAsyncAssertEqual(await eventIterator1.next(), .shutdownGracefully)
```
gjcairo, line 326: Shouldn't we also be getting a `runCancelled` event after the second service exits?

Sources/ServiceLifecycle/ServiceGroupConfiguration.swift
FranzBusch: Moved to use `successfulTerminationBehavior` and `failureTerminationBehavior`

Sources/UnixSignals/UnixSignalsSequence.swift
```diff
@@ -80,7 +81,8 @@ extension UnixSignalsSequence {
                 signal(sig.rawValue, SIG_IGN)
                 #endif
                 return .init(
-                    dispatchSource: DispatchSource.makeSignalSource(signal: sig.rawValue, queue: UnixSignalsSequence.queue),
+                    // This force-unwrap is safe since Dispatch always returns a `DispatchSource`
```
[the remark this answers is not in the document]
FranzBusch: Yeah and if we start to crash here then we will notice it and have to see what we get returned.

Tests/ServiceLifecycleTests/ServiceGroupTests.swift
FranzBusch: Good one but no we do not since a service returning when we are gracefully shutting down is expected.

gjcairo (approved):

Sources/ServiceLifecycle/ServiceGroupConfiguration.swift
```diff
@@ -41,17 +46,139 @@ public struct ServiceGroupConfiguration: Hashable, Sendable {
...
+                }
+            }
+        }
+
+        /// The service.
+        public var service: any Service
+        /// The behavior when the service returns from its `run()` method.
+        public var successfulTerminationBehavior: TerminationBehavior
+        /// The behavior when the service throws from its `run()` method.
+        public var failureTerminationBehavior: TerminationBehavior
```
glbrntt, line 80: mild nit: successful is an adjective, failure is a noun; can we have `success` instead of successful?

Sources/ServiceLifecycle/ServiceGroupConfiguration.swift
```diff
@@ -41,17 +46,139 @@ public struct ServiceGroupConfiguration: Hashable, Sendable {
...
+
+        /// Initializes a new ``ServiceGroupConfiguration/ServiceConfiguration``.
+        ///
+        /// - Parameters:
+        ///   - service: The service.
+        ///   - returnBehavior: The behavior when the service returns from its `run()` method.
+        ///   - throwBehavior: The behavior when the service throws from its `run()` method.
```
glbrntt, line 87: These are out-of-date

Sources/ServiceLifecycle/ServiceGroupConfiguration.swift
```diff
@@ -41,17 +46,139 @@ public struct ServiceGroupConfiguration: Hashable, Sendable {
...
+        cancellationSignals: [UnixSignal] = [],
+        logger: Logger
+    ) {
+        self.services = Array(services.map { ServiceConfiguration(service: $0) })
+        self.logger = logger
+        self.gracefulShutdownSignals = gracefulShutdownSignals
+        self.cancellationSignals = cancellationSignals
+    }
+
+    @available(*, deprecated)
```
glbrntt, line 178: Will this be removed before 2.0?

FranzBusch: We probably should.

glbrntt (approved):
LGTM!

PR #150, opened 2023-08-22
Clarify log statement around graceful shutdown

FranzBusch:


glbrntt (approved):

Sources/ServiceLifecycle/ServiceGroup.swift
```diff
@@ -275,7 +275,7 @@ public actor ServiceGroup: Sendable {
 
                     case .gracefullyShutdownGroup:
                         self.logger.error(
-                            "Service finished unexpectedly. Gracefully shutting down group.",
+                            "Service finished. Gracefully shutting down group.",
```
vlm, line 278: self.logger.debug?

PR #152, opened 2023-08-22
Propagate service error when gracefully shutting down

FranzBusch:
# Motivation

Propagate the service error when we shutdown the group after the service threw.

### DIFF
--- a/Sources/ServiceLifecycle/ServiceGroup.swift
+++ b/Sources/ServiceLifecycle/ServiceGroup.swift
@@ -309,25 +309,25 @@ public actor ServiceGroup: Sendable {
                         }
                     }
 
-                case .serviceThrew(let service, let index, let error):
+                case .serviceThrew(let service, let index, let serviceError):
                     switch service.failureTerminationBehavior.behavior {
                     case .cancelGroup:
                         self.logger.debug(
                             "Service threw error. Cancelling group.",
                             metadata: [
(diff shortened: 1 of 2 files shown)


Sources/ServiceLifecycle/ServiceGroup.swift
```diff
@@ -338,6 +338,7 @@ public actor ServiceGroup: Sendable {
                                 group: &group,
                                 gracefulShutdownManagers: gracefulShutdownManagers
                             )
+                            return .failure(error)
```
vlm, line 341: 🙇

Sources/ServiceLifecycle/ServiceGroup.swift
```diff
@@ -338,6 +338,7 @@ public actor ServiceGroup: Sendable {
...
                         } catch {
                             return .failure(error)
```
vlm, line 343: This is a different error, no?

vlm (approved):

Sources/ServiceLifecycle/ServiceGroup.swift
FranzBusch: Yes this is the error if the group fails to shutdown. What do you expect here? The error from the original service that threw or the error during shutdown?

FranzBusch: Agreed. We are also logging that another service threw during shutdown and what error it threw.

PR #158, opened 2023-09-11
Add async gracefulShutdown function to wait on graceful shutdown trigger

gjcairo:
# Motivation
Sometimes, it's useful to wait for graceful shutdown to happen before resuming execution on a task.

This can be of particular interest when we don't have anything to do in a Service's `run` method but want to make sure some cleanup happens on graceful shutdown.
In this scenario, we can achieve this by calling `await gracefulShutdown()` in `run()`, followed by whatever cleanup we must do.

# Modification
This PR adds a new `gracefulShutdown()` function to the `GracefulShutdown` suite. 

# Result
A new `gracefulShutdown()` method in the `GracefulShutdown` suite.

### DIFF
--- a/Sources/ServiceLifecycle/GracefulShutdown.swift
+++ b/Sources/ServiceLifecycle/GracefulShutdown.swift
@@ -55,6 +55,16 @@ public func withGracefulShutdownHandler<T>(
     return try await operation()
 }
 
+/// Waits until graceful shutdown is triggered.
+///
+/// This method suspends the caller until graceful shutdown is triggered. If the calling task is cancelled before
+/// graceful shutdown is triggered then this method will throw a `CancellationError`.
+///
+/// - Throws: `CancellationError` if the task is cancelled.
+public func gracefulShutdown() async throws {
+    try await AsyncGracefulShutdownSequence().first { _ in true }
+}
+
 /// This is just a helper type for the result of our task group.
 enum ValueOrGracefulShutdown<T: Sendable>: Sendable {
     case value(T)
@@ -72,7 +82,7 @@ public func cancelOnGracefulShutdown<T: Sendable>(_ operation: @Sendable @escapi
         }
 
         group.addTask {
-            for await _ in AsyncGracefulShutdownSequence() {
+            for try await _ in AsyncGracefulShutdownSequence() {
                 return .gracefulShutdown
             }
 
@@ -138,6 +148,8 @@ public final class GracefulShutdownManager: @unchecked Sendable {
         fileprivate var handlerCounter: UInt64 = 0
         /// A boolean indicating if we have been shutdown already.
         fileprivate var isShuttingDown = false
+        /// Continuations to resume after all of the handlers have been executed.
+        fileprivate var gracefulShutdownFinishedContinuations = [CheckedContinuation<Void, Never>]()
     }
 
     private let state = LockedValueBox(State())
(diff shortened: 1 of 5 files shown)


Sources/ServiceLifecycle/GracefulShutdown.swift
```diff
@@ -55,6 +55,18 @@ public func withGracefulShutdownHandler<T>(
     return try await operation()
 }
 
+/// Suspend the current task until graceful shutdown is executed.
+///
+/// A common use-case is to call this method in the `operation` closure of ``withGracefulShutdownHandler(operation:onGracefulShutdown:)``
+/// when you are only interested in executing a closure upon graceful shutdown, and nothing else.
```
FranzBusch, line 61: I think we should slightly change this
```suggestion
/// Waits until graceful shutdown is triggered.
///
/// This method suspends the caller until graceful shutdown is triggered. If the calling task is cancelled before
/// graceful shutdown is triggered then this method will exit with a `CancellationError`.
```

Sources/ServiceLifecycle/GracefulShutdown.swift
```diff
@@ -55,6 +55,18 @@ public func withGracefulShutdownHandler<T>(
...
+public func waitForGracefulShutdown() async {
```
FranzBusch, line 62: I think we should name this just `gracefulShutdown()` because you already call it with `await gracefulShutdown()`. This would be similar to what we have in XCTest here: https://developer.apple.com/documentation/xctest/xctestcase/4109476-fulfillment

Sources/ServiceLifecycle/GracefulShutdown.swift
```diff
@@ -55,6 +55,18 @@ public func withGracefulShutdownHandler<T>(
...
+    guard let gracefulShutdownManager = TaskLocals.gracefulShutdownManager else {
+        return
+    }
+    
+    await gracefulShutdownManager.waitForGracefulShutdown()
```
FranzBusch, line 67: That's not the right way to do this and it doesn't respect cancellation. We have a `AsyncGracefulShutdownSequence` that we should just use here with `try await AsyncGracefulShutdownSequence().first { _ in }`. 

Furthermore, we should probably change the implementation of that sequence to use an `AsyncThrowingStream` instead so we get the `CancellationError` thrown.

Tests/ServiceLifecycleTests/GracefulShutdownTests.swift
```diff
@@ -253,4 +253,23 @@ final class GracefulShutdownTests: XCTestCase {
             XCTAssertTrue(Task.isShuttingDownGracefully)
         }
     }
+    
+    func testWaitForGracefulShutdown() async throws {
```
FranzBusch, line 257: Can we have two more tests here:
1. Where graceful shutdown is already triggered before calling `await gracefulShutdown()`
2. Where we cancel while awaiting the graceful shutdown

FranzBusch (changes requested):

Sources/ServiceLifecycle/GracefulShutdown.swift
gjcairo: You think it's not worth mentioning the use-case scenario? I feel like we should mention this somewhere (although maybe the docs for `withGracefulShutdownHandler(operation:onGracefulShutdown:)` are a better place). 
Right now, we don't mention that the graceful shutdown handler won't be executed if the operation finished running, which may or may not be obvious to everyone.

FranzBusch (approved):
LGTM!

FranzBusch:

FranzBusch:

FranzBusch:

FranzBusch:

gjcairo:

gjcairo:

gjcairo:

gjcairo:

PR #162, opened 2023-10-13
Add a configurable escalation behaviour

FranzBusch:
# Motivation
The service group is often running multiple services and orchestrates shutdown for them. We have seen that sometimes some services never shutdown nor do they respond to task cancellation properly. This can become quite problematic when the whole application is waiting for a service to cancel and otherwise appears to be healthy but in reality can't serve any traffic.

# Modification
This PR adds a new configuration to escalate both graceful shutdown and cancellation. The escalation order is graceful shutdown -> task cancellation -> `fatalError`. The `fatalError` acts a last resort to make sure applications are never stuck.

### DIFF
--- a/Sources/ServiceLifecycle/ServiceGroup.swift
+++ b/Sources/ServiceLifecycle/ServiceGroup.swift
@@ -61,7 +61,7 @@ public actor ServiceGroup: Sendable {
         self.cancellationSignals = configuration.cancellationSignals
         self.logger = configuration.logger
         self.loggingConfiguration = configuration.logging
-        self.maximumGracefulShutdownDuration = configuration._maximumCancellationDuration
+        self.maximumGracefulShutdownDuration = configuration._maximumGracefulShutdownDuration
         self.maximumCancellationDuration = configuration._maximumCancellationDuration
     }
 
@@ -100,7 +100,7 @@ public actor ServiceGroup: Sendable {
         self.cancellationSignals = configuration.cancellationSignals
         self.logger = logger
         self.loggingConfiguration = configuration.logging
-        self.maximumGracefulShutdownDuration = configuration._maximumCancellationDuration
+        self.maximumGracefulShutdownDuration = configuration._maximumGracefulShutdownDuration
         self.maximumCancellationDuration = configuration._maximumCancellationDuration
     }
 
@@ -497,14 +497,13 @@ public actor ServiceGroup: Sendable {
 
         if #available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *), let maximumGracefulShutdownDuration = self.maximumGracefulShutdownDuration {
             group.addTask {
-                try await Task.sleep(for: Duration(
+                try? await Task.sleep(for: Duration(
                     secondsComponent: maximumGracefulShutdownDuration.secondsComponent,
                     attosecondsComponent: maximumGracefulShutdownDuration.attosecondsComponent
                 ))
                 return .gracefulShutdownTimedOut
             }
         }
-
 
         // We are storing the first error of a service that threw here.
         var error: Error?
@@ -657,6 +656,9 @@ public actor ServiceGroup: Sendable {
     ) {
         guard cancellationTimeoutTask == nil else {
             // We already have a cancellation timeout task running.
+            self.logger.debug(
+                "Task cancellation timeout task already running."
+            )
             return
         }
         group.cancelAll()
--- a/Sources/ServiceLifecycle/ServiceGroup.swift
+++ b/Sources/ServiceLifecycle/ServiceGroup.swift
@@ -33,6 +33,8 @@ public actor ServiceGroup: Sendable {
     private let logger: Logger
     /// The logging configuration.
     private let loggingConfiguration: ServiceGroupConfiguration.LoggingConfiguration
+    /// The escalation configuration.
+    private let escalationConfiguration: ServiceGroupConfiguration.EscalationBehaviour
     /// The signals that lead to graceful shutdown.
     private let gracefulShutdownSignals: [UnixSignal]
     /// The signals that lead to cancellation.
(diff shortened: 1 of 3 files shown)


Sources/ServiceLifecycle/ServiceGroup.swift
```diff
@@ -267,6 +277,13 @@ public actor ServiceGroup: Sendable {
                 }
             }
 
+            group.addTask {
+                // This child task is waiting forever until the group gets cancelled.
+                let (stream, _) = AsyncStream.makeStream(of: Void.self)
+                await stream.first { _ in true }
```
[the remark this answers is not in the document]
FranzBusch: But what do you put in? `Task.sleep(nanoseconds: .max)` still has the potential to run out. Though realistically it will never.

Sources/ServiceLifecycle/ServiceGroupConfiguration.swift
```diff
@@ -96,6 +96,65 @@ public struct ServiceGroupConfiguration: Sendable {
...
+                if let newValue = newValue {
+                    self._maximumCancellationDuration = (newValue.components.seconds, newValue.components.attoseconds)
+                } else {
+                    self._maximumCancellationDuration = nil
+                }
+            }
+        }
+
+        private var _maximumGracefulShutdownDuration: (secondsComponent: Int64, attosecondsComponent: Int64)?
+        private var _maximumCancellationDuration: (secondsComponent: Int64, attosecondsComponent: Int64)?
```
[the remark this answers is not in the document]
FranzBusch: It is not allowed to add availability on a stored property. So we have to do this dance somewhere.

Sources/ServiceLifecycle/ServiceGroupConfiguration.swift
```diff
@@ -96,6 +96,65 @@ public struct ServiceGroupConfiguration: Sendable {
         }
     }
 
+    /// The group's escalation configuration.
+    public struct EscalationBehaviour: Sendable {
```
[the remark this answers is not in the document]
FranzBusch: Agreed. I removed the struct and just made them top-level properties

Sources/ServiceLifecycle/ServiceGroup.swift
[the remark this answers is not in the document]
glbrntt: `Int64.max` nanoseconds is ~580 years so realistically not an issue. I'm totally fine with the async stream, I just wouldn't have thought to use it!

Sources/ServiceLifecycle/ServiceGroup.swift
```diff
@@ -584,6 +650,43 @@ public actor ServiceGroup: Sendable {
...
+        group: inout ThrowingTaskGroup<ChildTaskResult, Error>,
+        cancellationTimeoutTask: inout Task<Void, Never>?
+    ) {
+        guard cancellationTimeoutTask == nil else {
+            // We already have a cancellation timeout task running.
+            return
+        }
+        group.cancelAll()
+
+
```
glbrntt, line 664:
```suggestion

```

Sources/ServiceLifecycle/ServiceGroup.swift
```diff
@@ -584,6 +650,43 @@ public actor ServiceGroup: Sendable {
...
+            // from being cancelled.
+            cancellationTimeoutTask = Task {
+                do {
+                    self.logger.debug(
+                        "Task cancellation timeout task started."
+                    )
+                    try await Task.sleep(for: Duration(
+                        secondsComponent: maximumCancellationDuration.secondsComponent,
+                        attosecondsComponent: maximumCancellationDuration.attosecondsComponent
+                    ))
```
glbrntt, line 677: `maximumCancellationDuration` is already a `Duration` here, I think?

Sources/ServiceLifecycle/ServiceGroup.swift
```diff
@@ -33,6 +33,10 @@ public actor ServiceGroup: Sendable {
     private let logger: Logger
     /// The logging configuration.
     private let loggingConfiguration: ServiceGroupConfiguration.LoggingConfiguration
+    /// The maximum amount of time that graceful shutdown is allowed to take.
+    private var maximumGracefulShutdownDuration: (secondsComponent: Int64, attosecondsComponent: Int64)?
+    /// The maximum amount of time that task cancellation is allowed to take.
+    private var maximumCancellationDuration: (secondsComponent: Int64, attosecondsComponent: Int64)?
```
glbrntt, line 39: Why are these mutable? Also complete nit, but you'll save some repetition converting between the tuples and `Duration` if you define your own `_Duration`.

Sources/ServiceLifecycle/ServiceGroup.swift
```diff
@@ -447,18 +481,31 @@ public actor ServiceGroup: Sendable {
...
         guard case .running = self.state else {
             fatalError("Unexpected state")
         }
 
+        if #available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *), let maximumGracefulShutdownDuration = self.maximumGracefulShutdownDuration {
+            group.addTask {
+                try await Task.sleep(for: Duration(
+                    secondsComponent: maximumGracefulShutdownDuration.secondsComponent,
+                    attosecondsComponent: maximumGracefulShutdownDuration.attosecondsComponent
+                ))
```
glbrntt, line 503: I think `maximumGracefulShutdownDuration` is already a `Duration` here, no need to convert again

glbrntt (approved):
Looks good modulo nits

Sources/ServiceLifecycle/ServiceGroup.swift
FranzBusch: No it's not. We are storing the raw components again because we cannot store a `Duration` due to the availability

Sources/ServiceLifecycle/ServiceGroup.swift
FranzBusch: Good catch with the `var`. Was a bad copy and paste. I decided not to introduce a custom type for now. if we do more `Duration` based APIs we should come back to that.

PR #167, opened 2023-12-04
Adopt `AsyncAlgorithms`

FranzBusch:
# Motivation

Now that `swift-async-algorithms` has a `1.0.0` tag we can depend on it instead of shipping our own version of `merge`.


tomerd:

tomerd (approved):

PR #170, opened 2024-01-24
Make sure we return after gracefully shutdown

FranzBusch:
# Motivation

After calling `shutdownGracefully` in our `ServiceGroup` we know that one of two things happened:
1. We got an error so something went wrong
2. We got no error so all services finished

In the 2. case, there might still be a few child tasks in the task group e.g. the graceful shutdown timeout task. We are currently still iterating the remaining child task results which might lead into unexpected fatal errors.

# Modification
This PR makes sure that after we have successfully shutdown we are stopping iterating the child task results.

# Result
No more unexpected fatal errors.


gjcairo (approved):

FranzBusch:

PR #171, opened 2024-01-26
Fix logic issues during graceful shutdown

FranzBusch:
# Motivation

This should fix the remaining issues raised in https://github.com/swift-server/swift-service-lifecycle/pull/166. The problem here was that if a service finished/threw out of order then we were wrongly treating this as if the service that we are currently shutting down finished.

# Modification
This PR ensures that we use the same `services` array during the graceful shutdown to nil out services that have finished. This way we correctly keep track of any service that finished. Additionally, there was a separate bug where we started to shutdown the next service to early if another service threw and had the termination behaviour of `shutdownGracefully`.

# Result
No more incorrect shutdown orderings.

### DIFF
--- a/Sources/ServiceLifecycle/ServiceGroup.swift
+++ b/Sources/ServiceLifecycle/ServiceGroup.swift
@@ -327,7 +327,7 @@ public actor ServiceGroup: Sendable {
                         services[index] = nil
                         do {
                             try await self.shutdownGracefully(
-                                services: services,
+                                services: &services,
                                 cancellationTimeoutTask: &cancellationTimeoutTask,
                                 group: &group,
                                 gracefulShutdownManagers: gracefulShutdownManagers
@@ -380,7 +380,7 @@ public actor ServiceGroup: Sendable {
 
                         do {
                             try await self.shutdownGracefully(
-                                services: services,
+                                services: &services,
                                 cancellationTimeoutTask: &cancellationTimeoutTask,
                                 group: &group,
                                 gracefulShutdownManagers: gracefulShutdownManagers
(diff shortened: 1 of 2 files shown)


Sources/ServiceLifecycle/ServiceGroup.swift
```diff
@@ -587,8 +589,15 @@ public actor ServiceGroup: Sendable {
                             error = serviceError
                         }
 
-                        // We can continue shutting down the next service now
-                        continue gracefulShutdownLoop
+                        if index == gracefulShutdownIndex {
+                            // The service that we were shutting down right now threw. Since it's failure
```
gjcairo, line 593: Nit
```suggestion
                            // The service that we were shutting down right now threw. Since its failure
```

Sources/ServiceLifecycle/ServiceGroup.swift
```diff
@@ -587,8 +589,15 @@ public actor ServiceGroup: Sendable {
...
+                            // behaviour is to shutdown the group we can continue
+                            continue gracefulShutdownLoop
+                        } else {
+                            // Another service threw while we were waiting for a shutdown
+                            // We have to continue the iterating the task group's result
```
gjcairo, line 598: Nit
```suggestion
                            // We have to continue iterating the task group's result
```

Tests/ServiceLifecycleTests/ServiceGroupTests.swift
```diff
@@ -943,7 +943,8 @@ final class ServiceGroupTests: XCTestCase {
             // Let's throw from the second service
             await service2.resumeRunContinuation(with: .failure(ExampleError()))
 
-            // The first service should still be running
+            // The first service should still be running but seeing a graceful shutdown signal firsts
```
gjcairo, line 946:
```suggestion
            // The first service should still be running but seeing a graceful shutdown signal first
```

gjcairo (approved):
LGTM, just some tiny doc nits.

PR #173, opened 2024-02-11
fix ambiguous use of init

tib:
The following code snippet emits an 'Ambigous use of init' error when using Swift Service Lifecycle v2.3.0:

```swift
let config = ServiceGroupConfiguration(
    services: [],
    logger: logger
)
```

This problem can be hotfixed like this, but it's not the right way:

```swift
let services: [ServiceGroupConfiguration.ServiceConfiguration] = []
let config = ServiceGroupConfiguration(
    services: services,
    logger: logger
)
```

This PR aims to fix the underlying issue in the library itself by removing the unnecessary `init` method, that caused the compiler error.

### DIFF
--- a/Sources/ServiceLifecycle/ServiceGroupConfiguration.swift
+++ b/Sources/ServiceLifecycle/ServiceGroupConfiguration.swift
@@ -171,19 +171,6 @@ public struct ServiceGroupConfiguration: Sendable {
     ///
     /// - Parameters:
     ///   - services: The groups's service configurations.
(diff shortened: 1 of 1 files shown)


FranzBusch:
Ah that's an unfortunate unambiguity but the proposed change here would be breaking API. In hindsight we should have made the parameter `serviceConfigurations` and `services` for the other one. What we could do is a deprecation dance but I am unsure if that's worth it. How many people are actually passing an empty array to a `ServiceGroup`?

PR #174, opened 2024-02-15
Make cancelOnGracefulShutdown operations nonescaping

tannerdsilva:
Hello.

As of release `v2.4.0`, please consider the following function from the public API.

```
public func cancelOnGracefulShutdown<T: Sendable>(_ operation: @Sendable @escaping () async throws -> T) async rethrows -> T?
```

I was working on a new project today and while architecting the shutdown procedures, I found myself wanting to use this function a lot. Very useful!

Unfortunately, I'm the kind of programmer that gets hopelessly lost in the details and semantics, and this is one of those scenarios where a perfectly valid implementation sits with me as "sub-optimal". Why? Well, I don't really see why this needs to take an escaping operation. In my personal utopian world, this function would work exactly as it is implemented today, simply with a **_non-escaping operation arg_**.

As we know, the primary factor that distinguishes an escaping block and non-escaping block is _when_ the function will be executed. In order to be considered non-escaping, the operation **must** execute within the lifecycle of the function. To my best interpretation, `cancelOnGracefulShutdown` looks to abide by this requirement, despite the compiler being unable to "see" this in the content. Indeed, it appears like this `operation` arg **always** executes within the boundaries of the root function.

As such, I have implemented this proposal to force the nonescaping semantics on this function. I'm no expert of compilers by any stretch of the imagination, but I suspect the compiler will be able to work much more efficiently with memory with the correct knowledge that the `operation` never escapes.

### DIFF
--- a/Sources/ServiceLifecycle/GracefulShutdown.swift
+++ b/Sources/ServiceLifecycle/GracefulShutdown.swift
@@ -74,42 +74,44 @@ enum ValueOrGracefulShutdown<T: Sendable>: Sendable {
 /// Cancels the closure when a graceful shutdown was triggered.
 ///
 /// - Parameter operation: The actual operation.
-public func cancelOnGracefulShutdown<T: Sendable>(_ operation: @Sendable @escaping () async throws -> T) async rethrows -> T? {
-    return try await withThrowingTaskGroup(of: ValueOrGracefulShutdown<T>.self) { group in
-        group.addTask {
-            let value = try await operation()
-            return .value(value)
-        }
-
-        group.addTask {
-            for try await _ in AsyncGracefulShutdownSequence() {
-                return .gracefulShutdown
-            }
-
-            throw CancellationError()
-        }
-
-        let result = try await group.next()
-        group.cancelAll()
-
-        switch result {
-        case .value(let t):
-            return t
-        case .gracefulShutdown:
-            switch try await group.next() {
-            case .value(let t):
-                return t
-            case .gracefulShutdown:
-                fatalError("Unexpectedly got gracefulShutdown from group.next()")
-
-            case nil:
-                fatalError("Unexpectedly got nil from group.next()")
-            }
-
-        case nil:
-            fatalError("Unexpectedly got nil from group.next()")
-        }
-    }
+public func cancelOnGracefulShutdown<T:Sendable>(_ operation: @Sendable () async throws -> T) async rethrows -> T? {
+	return try await withoutActuallyEscaping(operation, do: { operationEscaping in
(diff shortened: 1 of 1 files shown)


Sources/ServiceLifecycle/GracefulShutdown.swift
```diff
@@ -74,42 +74,44 @@ enum ValueOrGracefulShutdown<T: Sendable>: Sendable {
...
-            case nil:
-                fatalError("Unexpectedly got nil from group.next()")
-            }
-
-        case nil:
-            fatalError("Unexpectedly got nil from group.next()")
-        }
-    }
+public func cancelOnGracefulShutdown<T:Sendable>(_ operation: @Sendable () async throws -> T) async rethrows -> T? {
+	return try await withoutActuallyEscaping(operation, do: { operationEscaping in
```
FranzBusch, line 78: This is currently not safe to do see: https://forums.swift.org/t/pitch-non-escapable-types-and-lifetime-dependency/69865/23

Once `TaskGroup` becomes `~Escapable` we can make this change here.

tannerdsilva: Thanks for the info. Hadn't seen this - will definitely follow this development and discussion.

sliemeobn: @FranzBusch how sure are we that we cannot deem this "safe enough"?
the `TaskGroup` escaping would be problematic anyway, and in "normal use" (and the function is short enough to scan that it is) a non-escaping closure should be fine, right?

IIUC John McCall confirmed [here](https://forums.swift.org/t/is-using-withoutactuallyescaping-within-the-context-of-a-taskgroup-valid/74038/9) that usage like this is fine.

FranzBusch: Just coming back to this. I am very hesitant to make that change even though it might be fine in theory. Do we have a real motivating example where the current API is limiting?

sliemeobn: I had a case: a "never-ending" SSE stream in a hummingbird response writer closure thing.
if I remember correctly, the stream writer is an `inout` parameter, so I could not wrap the producing loop it in a `cancelOnGracefulShutdown` block.

in my case I was able to rework the logic to be `AsyncSequence`-based and use `for await ... in myStream.cancelOnGracefulShutdown()` - maybe even cleaner, but it still felt like a real limitation.

FranzBusch: I see. That's really unfortunate. Let's keep an eye on this and see if more such problems happen and then we can re-evaluate if we should continue with this here.

PR #176, opened 2024-02-16
Add `withTaskCancellationOrGracefulShutdownHandler`

adam-fowler:
Execute an operation with a graceful shutdown or task cancellation handler that’s immediately invoked if the current task is shutting down gracefully or has been cancelled.

### DIFF
--- a/Tests/ServiceLifecycleTests/GracefulShutdownTests.swift
+++ b/Tests/ServiceLifecycleTests/GracefulShutdownTests.swift
@@ -303,4 +303,54 @@ final class GracefulShutdownTests: XCTestCase {
             XCTAssertTrue(error is CancellationError)
         }
     }
+
+    func testWithTaskCancellationOrGracefulShutdownHandler_GracefulShutdown() async throws {
+        var cont: AsyncStream<Void>.Continuation!
+        let stream = AsyncStream<Void> { cont = $0 }
+        let continuation = cont!
+
+        await testGracefulShutdown { gracefulShutdownTestTrigger in
+            await withTaskCancellationOrGracefulShutdownHandler {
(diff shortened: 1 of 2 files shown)


Tests/ServiceLifecycleTests/GracefulShutdownTests.swift
```diff
@@ -303,4 +303,54 @@ final class GracefulShutdownTests: XCTestCase {
...
+
+                    await group.waitForAll()
+                }
+            } onCancelOrGracefulShutdown: {
+                continuation.finish()
+            }
+        }
+    }
+
+    func testWithTaskCancellationOrGracefulShutdownHandler_TaskCancellation() async throws {
```
adam-fowler, line 329: It was quite awkward to construct a test that actually tested cancellation as AsyncStream.next() exits immediately if  it is cancelled.

FranzBusch: I don't think you need any of the streams here. Just create a `LockedValueBox<CheckedContinuation<Void, Never>>` and then in the operation closure you create the continuation and store it into the lock and in the handler closure you just resume it.

adam-fowler: I need the first stream to ensure the withTaskCancellationOrGracefulShutdownHandler operation actually gets run. 

The locked value box would have to be a `LockedValueBox<CheckedContinuation<Void, Never>?>` and then if I was to be correct I would have to deal with the situation where the cancellation handler gets called before the operation and optional CheckedContinuation is still `nil`

FranzBusch (approved):

PR #177, opened 2024-02-21
`cancelOnGracefulShutdown` hangs, if cancellation is not immediately

fabianfett:
### Motivation

Currently `cancelOnGracefulShutdown` hangs forever, if cancellation is not immediately but returns a eventually successful result (test case 1: `testCancelOnGracefulShutdownSurvivesCancellation`).
If the Cancellation leads to another error than `CancellationError`, this error is currently not propagated to the user (test case 2: `testCancelOnGracefulShutdownSurvivesErrorThrown`).

### Changes

- `AsyncGracefulShutdownSequence` should never throw as this hides the real errors otherwise
- `CancellationWaiter` should never throw as this hides the real errors otherwise
- Adjustments in `cancelOnGracefulShutdown`

### Result

- `cancelOnGracefulShutdown` works as expected

### DIFF
--- a/Sources/ServiceLifecycle/AsyncGracefulShutdownSequence.swift
+++ b/Sources/ServiceLifecycle/AsyncGracefulShutdownSequence.swift
@@ -20,7 +20,7 @@
 @usableFromInline
 struct AsyncGracefulShutdownSequence: AsyncSequence, Sendable {
     @usableFromInline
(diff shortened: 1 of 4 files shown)


FranzBusch (approved):
Nice fix! Thanks for doing the work.

Issue #179, opened 2024-04-17
Question: return type of `cancelOnGracefulShutdown`


sidepelican:
https://github.com/swift-server/swift-service-lifecycle/blob/f5780d95bdf3f740f14decb6c8c823815c883451/Sources/ServiceLifecycle/GracefulShutdown.swift#L118
I noticed that the `cancelOnGracefulShutdown` method signature suggests it returns an optional (`T?`). 
However, I couldn’t find any code paths in the implementation where a `nil` value would be returned.

Could you please clarify why the return type is optional in this context?
Is there a scenario where this method might return `nil` that I might have missed?

FranzBusch:
You are right this seems like a mistake we should fix. This method indeed shouldn't return an optional. Would you be open to file a PR for this?

PR #181, opened 2024-04-24
Global cancelOnGracefulShutdown returns non nil value

sidepelican:
Issue: https://github.com/swift-server/swift-service-lifecycle/issues/179

`cancelOnGracefulShutdown` has no code paths that return a `nil` value.

### DIFF
--- a/Sources/ServiceLifecycle/GracefulShutdown.swift
+++ b/Sources/ServiceLifecycle/GracefulShutdown.swift
@@ -115,7 +115,7 @@ enum ValueOrGracefulShutdown<T: Sendable>: Sendable {
 /// Cancels the closure when a graceful shutdown was triggered.
 ///
 /// - Parameter operation: The actual operation.
-public func cancelOnGracefulShutdown<T: Sendable>(_ operation: @Sendable @escaping () async throws -> T) async rethrows -> T? {
+public func cancelOnGracefulShutdown<T: Sendable>(_ operation: @Sendable @escaping () async throws -> T) async rethrows -> T {
     return try await withThrowingTaskGroup(of: ValueOrGracefulShutdown<T>.self) { group in
         group.addTask {
             let value = try await operation()
--- a/Sources/ServiceLifecycle/GracefulShutdown.swift
+++ b/Sources/ServiceLifecycle/GracefulShutdown.swift
@@ -115,7 +115,7 @@ enum ValueOrGracefulShutdown<T: Sendable>: Sendable {
 /// Cancels the closure when a graceful shutdown was triggered.
 ///
 /// - Parameter operation: The actual operation.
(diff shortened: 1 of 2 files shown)


Sources/ServiceLifecycle/GracefulShutdown.swift
```diff
@@ -115,7 +115,7 @@ enum ValueOrGracefulShutdown<T: Sendable>: Sendable {
 /// Cancels the closure when a graceful shutdown was triggered.
 ///
 /// - Parameter operation: The actual operation.
-public func cancelOnGracefulShutdown<T: Sendable>(_ operation: @Sendable @escaping () async throws -> T) async rethrows -> T? {
+public func cancelOnGracefulShutdown<T: Sendable>(_ operation: @Sendable @escaping () async throws -> T) async rethrows -> T {
```
FranzBusch, line 118: So we gotta be a bit careful here since this is a breaking change. Can we deprecate this method and introduce the same one. This one can call the new one to avoid duplication.

FranzBusch (changes requested):

Sources/ServiceLifecycle/GracefulShutdown.swift
sidepelican: I was also concerned about that, but couldn't find a suitable solution.

We could provide an overload method without an ambiguous error as follows:

```swift
@available(*, deprecated)
@_disfavoredOverload
public func cancelOnGracefulShutdown<T: Sendable>(_ operation: @Sendable @escaping () async throws -> T) async rethrows -> T? {
    ...
}

public func cancelOnGracefulShutdown<T: Sendable>(_ operation: @Sendable @escaping () async throws -> T) async rethrows -> T {
    ...
}
```

However, this overload does not address the source-breaking issue.
Since `T` can be assigned to `T?`, the new method will be used automatically.

```swift
// OK
let value: Int? = await cancelOnGracefulShutdown {
    return 42
}
```

Nevertheless, if there were a use case like the following, it would break existing code, and we couldn't resolve it with an overload:

```swift
// Source breaking!
let value = await cancelOnGracefulShutdown {
    return 42
}
if let value {
    ...
}
```

Aside from providing an overload, the only other option would be to use a different name, but I didn’t think it was appropriate to rename this function.

I just can't think of a good approach.

FranzBusch: Yep that's completely true. We gotta introduce a new name here sadly. Let's deprecate the current method and introduce a new one called `cancelWhenGracefulShutdown`. It isn't great but it gets us moving forward.

sidepelican: OK, I have implemented `cancelWhenGracefulShutdown`.

FranzBusch (approved):
Thanks!

FranzBusch:

PR #183, opened 2024-05-29
Stop using renamed pattern which can cause compiler crash

PeterAdams-A:
Motivation:

Deprecation with rename fixit has been seen to cause compiler crashes in 5.x compiler versions.

Modifications:

Deprecate with message for 5.x compilers.

Result:

Compilation without failure in all scenarios


gjcairo (approved):

gjcairo:

FranzBusch (approved):

PeterAdams-A:
https://github.com/apple/swift/pull/68053 I believe is the fix to the compiler.

PR #184, opened 2024-06-02
add support for trapping sigabrt, sigill, sigsegv

zg:
- sigabrt: usually generated by the abort() function. Useful for cleanup prior to termination.
- sigill: issued if the user attempts to execute an illegal, malformed, or privileged instruction.
- sigsegv: issued if the user makes an invalid memory reference, such as dereferencing a null or invalid pointer.


FranzBusch (approved):
LGTM!

PR #189, opened 2024-11-22
Make logging optional and update docs

nervenes:
Hello! :wave: This PR contains a few changes, notably:

- Updating the ``README.md`` to use the new initializer instead of the deprecated one.
- Add a ``renamed`` to the deprecated initializer.
- Updating the docs to use the simpler initializer.
- Make the ``logging`` parameter on the initializers optional, the motivation for this is in cases the user don't or can't do logging, they don't have to construct a logger and bootstrap the logging system with ``SwiftLogNoOpLogHandler``, so this code:
```swift
LoggingSystem.bootstrap(SwiftLogNoOpLogHandler.init)
let serviceGroup = ServiceGroup(
    services: [service],
    cancellationSignals: [.sigint],
    logger: .init(label: "")
)
```
becomes
```swift
let serviceGroup = ServiceGroup(
    services: [service],
    cancellationSignals: [.sigint]
)
```
and it also allows the user to do logging outside the ServiceGroup without necessarily doing logging within it.

If there is anything to improve, let me know! :smiley:

*thanks dimi and gwynne for helping along the way and answering my questions :)*

### DIFF
--- a/Sources/ServiceLifecycle/ServiceGroupConfiguration.swift
+++ b/Sources/ServiceLifecycle/ServiceGroupConfiguration.swift
@@ -172,10 +172,14 @@ public struct ServiceGroupConfiguration: Sendable {
     ///   - logger: The group's logger.
     public init(
         services: [ServiceConfiguration],
-        logger: Logger
+        logger: Logger? = nil
     ) {
+        if logger == nil {
+            LoggingSystem.bootstrap { SwiftLogNoOpLogHandler($0) }
+        }
+        
         self.services = services
-        self.logger = logger
+        self.logger = logger ?? .init(label: "")
     }
 
     /// Initializes a new ``ServiceGroupConfiguration``.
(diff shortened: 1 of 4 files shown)


Sources/ServiceLifecycle/ServiceGroupConfiguration.swift
```diff
@@ -107,7 +107,7 @@ public struct ServiceGroupConfiguration: Sendable {
     public var cancellationSignals = [UnixSignal]()
 
     /// The group's logger.
-    public var logger: Logger
+    public var logger: Logger?
```
FranzBusch, line 110: We can't make this change since this an API break. I think at this point there is no nice way to make this optional anymore sadly.

nervenes: I see, I've reverted the changes of making it an optional, with only improved docs and readme examples remaining 👍

FranzBusch (approved):

FranzBusch:
Thanks for the change. Let's get this landed!

Sources/ServiceLifecycle/ServiceGroupConfiguration.swift
dimitribouniol: An alternative approach to make the logger in the initializers optional could be to chain the log handler from the passed in optional logger to the stored-property logger:
```swift
self.logger = Logger(logger.label ?? "ServiceLifecycle", factory: { _ in logger?.handler ?? SwiftLogNoOpLogHandler() })
```

PR #192, opened 2024-12-09
UnixSignal: Add SIGCONT

juri:
I'm using swift-service-lifecycle's UnixSignal in a text UI library, and needed SIGCONT. It's probably not super relevant in the service use case, but it looks like UnixSignal is the best way to handle signals, service or not, and I don't know if there's a downside to the library supporting it?


FranzBusch (approved):
Thanks. Happy to add this to our list of signals!

PR #197, opened 2025-01-12
Update README.md

FranzBusch:
The readme shows an actor implementing the `Service` protocol. This has lead to some people thinking that only actors can do this so let's switch this to a `class` which is the usual type conforming to `Service`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,8 @@ the configured signals and trigger a graceful shutdown on each service.
 import ServiceLifecycle
 import Logging
 
-final class FooService: Service {
+// A service can be implemented by a struct, class or actor. For this example we are using a struct.
+struct FooService: Service {
     func run() async throws {
         print("FooService starting")
         try await Task.sleep(for: .seconds(10))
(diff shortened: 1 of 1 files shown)


czechboy0:
Why not a struct? Most services I conform are actually either structs or actors, but rarely classes, as those then require manual locking.

gjcairo:
Yeah I kind of agree that `class` may also be misleading.

Regardless of whether we keep `class` or change it to `struct` though, can we not add a comment saying that it actually _doesn't matter_ whether the type is a `class`, an `actor` or a `struct`? If it's brought confusion in the past, let's make it super obvious.

README.md
```diff
@@ -66,7 +66,7 @@ the configured signals and trigger a graceful shutdown on each service.
 import ServiceLifecycle
 import Logging
 
-actor FooService: Service {
+final class FooService: Service {
```
FranzBusch, line 69:
```suggestion
// A service can be implemented by a struct, class or actor. For this example we are using a struct.
struct FooService: Service {
```

FranzBusch:
@czechboy0 @gjcairo updated the example

gjcairo (approved):

Issue #185, opened 2024-06-13
`ServiceGroup` should support dynamically adding `Services` while its running


fabianfett:
There are scenarios in which users need a running Service in order to retrieve data. This data shall then be used to spawn another Service.

An example for this could be querying AWS SSM for a secret that is then used inside a Lambda.

I imagine the code to look something like this:

```swift
import Soto
import SotoSSM

enum MyLambda {
    static func main() async throws {
        let awsClient = AWSClient()
        let ssmClient = SSMClient(awsClient)

        try await withTaskGroup { taskGroup in
            let serviceGroup = ServiceGroup([awsClient])
            taskGroup.addTask {
                try await serviceGroup.run()
            }

            let mySecret = try await ssmClient.getSecret()

            let runtime = LambdaRuntime { (event: Request, context) in
                // use mySecret
            }

            // add runtime to already running serviceGroup.
            serviceGroup.add(runtime)
        }
    }
}
```

FranzBusch:
I think we can totally do that and be able to add dynamic services to the group. I was briefly thinking if having dependency between services during startup could be used here. I think it could but it would become awkward to spell. So yeah great feature requests!

sliemeobn:
I was about to give this a go, but got stuck on a fundamental API question:

The inner plumbings are quite straight forward I would say (some async sequence of `ServiceConfiguration` to pipe new services into the main `run` loop and set them up there.)

I was going in with a non-throwing `nonisolated public func addService(...)`, but that would essentially mean: no access to the group's state, no feedback whatsoever if the group was even added or its run called.

```swift
public actor ServiceGroup {
    //...
    private let 
    nonisolated private let addedServices: AsyncStream<ServiceGroupConfiguration.ServiceConfiguration>.Continuation

    //...
    nonisolated public func addService(_ serviceConfiguration: ServiceGroupConfiguration.ServiceConfiguration) {
        // Without making `addService` throwing and/or async, there is not a lot we could do with the result of yield
        addedServices.yield(serviceConfiguration)
    }
```

And, you'd theoretically have an unbounded buffer of added services if nobody ever calls `run`.

The alternative is to access actor state in `addService`, resulting in an `async` and probably throwing function to signal things like "already finished". Cleaner in a way (back-pressure, error-channel for funky states, ...) but not exactly a friendly API to call...

Any thoughts?

czechboy0:
Hi @sliemeobn, 

> but not exactly a friendly API to call

[...]

PR #199, opened 2025-03-05
Support adding services while `ServiceGroup` is running

sliemeobn:
somewhat straight-forward implementation draft of 
https://github.com/swift-server/swift-service-lifecycle/issues/185

choices were made, what do we think?


Sources/ServiceLifecycle/ServiceGroup.swift
```diff
@@ -106,6 +108,37 @@ public actor ServiceGroup: Sendable, Service {
...
+        case var .initial(services: services):
+            self.state = .initial(services: [])
+            services.append(serviceConfiguration)
+            self.state = .initial(services: services)
+
+        case .running(_, let addedServiceChannel):
+            await addedServiceChannel.send(serviceConfiguration)
+
+        case .finished:
+            return
```
FranzBusch, line 128: I think this case definitely needs to throw an error to indicate something went wrong

Sources/ServiceLifecycle/ServiceGroup.swift
```diff
@@ -310,6 +333,49 @@ public actor ServiceGroup: Sendable, Service {
...
+
+                        _unownedTaskGroupHandledCarefully.addServiceTask(
+                            serviceConfiguration,
+                            gracefulShutdownManager: gracefulShutdownManager,
+                            index: services.count - 1
+                        )
+                    }
+                } onCancel: {
+                    addedServiceChannel.finish()
+                }
```
FranzBusch, line 367: This shouldn't be necessary since the channel setups up cancellation handlers as well and the for loop above will finish.

Sources/ServiceLifecycle/ServiceGroup.swift
```diff
@@ -310,6 +333,49 @@ public actor ServiceGroup: Sendable, Service {
...
+                        self.logger.debug(
+                            "Starting added service",
+                            metadata: [
+                                self.loggingConfiguration.keys.serviceKey: "\(serviceConfiguration.service)"
+                            ]
+                        )
+
+                        let gracefulShutdownManager = GracefulShutdownManager()
+                        gracefulShutdownManagers.append(gracefulShutdownManager)
+                        services.append(serviceConfiguration)
```
FranzBusch, line 352: I don't see how this is safe from a concurrency point of view since the services & gracefulShutdownManagers variables are in a different task. This is probably causing concurrency issues in Swift 6.

Sources/ServiceLifecycle/ServiceGroup.swift
```diff
@@ -310,6 +333,49 @@ public actor ServiceGroup: Sendable, Service {
                 "We did not create a graceful shutdown manager per service"
             )
 
+            var _unownedTaskGroupHandledCarefully = group
```
FranzBusch, line 336: I think we need to come up with a different solution here than unsafe escaping the task group into a child task.

FranzBusch (commented):
This is a great start. Thanks for working on this. I left some feedback inline.

Sources/ServiceLifecycle/ServiceGroup.swift
sliemeobn: ^^ I figured you'd say that.
I am pretty sure this is perfectly safe as written because we do it "withoutActuallyEscaping" if you will.

there is no good alternative I can think of would be not be either a convoluted merging of the two streams (ie: the group itself and the added services channel - which would then add to the group?) or a complete overhaul of the run loop and wind-down loop.

Sources/ServiceLifecycle/ServiceGroup.swift
sliemeobn: I am not sure I follow. isn't everything actor-isolated to the `ServiceGroup`?

Sources/ServiceLifecycle/ServiceGroup.swift
FranzBusch: If that's the case it sounds like a bug in the `AsyncChannel` to me

Sources/ServiceLifecycle/ServiceGroup.swift
FranzBusch: The child task isn't actor isolated since the closures of `addTask` are `sending`. Maybe I'm wrong but this looks to me like something that would result in an error in language mode 6. Could you check that real quick?

Sources/ServiceLifecycle/ServiceGroup.swift
FranzBusch: > I am pretty sure this is perfectly safe as written because we do it "withoutActuallyEscaping" if you will.
I agree that this is most likely safe but we can't prove it to the compiler sadly. Let me think a bit about it.

FranzBusch: Just after having written this I actually thought of a something where this isn't safe. The underlying task group implementation does in fact no guarantee that you can call `addTask` concurrently nor does the `group.next()` support concurrent iteration. The fact that this works is pure luck and nothing that we can expect to stay this way.

Sources/ServiceLifecycle/ServiceGroup.swift
sliemeobn: you are absolutely correct (I am not used to writing code without strict concurrency checking any more ; )
ok then, this settles the question about the escaping task group as well - we should funnel everything to the main loop on the actor somehow...

Sources/ServiceLifecycle/ServiceGroup.swift
sliemeobn: yeah, as mentioned above: after realizing that `.addTask` does not run on the actors isolation the whole thing falls apart anyway... we'll need a merged async sequence I think...

sliemeobn:
^^ I was just coding this exact thing, but I wasn't sure if it is too crazy ; )
I'll see how it goes...

sliemeobn:
@FranzBusch I managed to clean up the channel reading, I think that part is good enough now.

The remaining question is what to do with added services during shutdown or cancel?

If we want a reliable back-propagation of "has it run or not?" we'll need to pass along a continuation I think and we'd need to make sure to fully drain and actively "reject" all added service until we set the status to `.finished`.

A lot of brittle and racy signaling... if we simply ignored (like now) life would be a lot easier ; )

Sources/ServiceLifecycle/ServiceGroup.swift
FranzBusch: I think that's fair. Let's ignore my comment then. `addService` is a best effort thing and if the group is finished or shutting down it won't run the new service. I think we just need to add a debug log for when that happens.

Sources/ServiceLifecycle/ServiceGroup.swift
```diff
@@ -779,6 +847,43 @@ public actor ServiceGroup: Sendable, Service {
     }
 }
 
+extension ThrowingTaskGroup where Failure == Error, ChildTaskResult == ServiceGroup.ChildTaskResult {
```
FranzBusch, line 850: This is more of a nit but I would prefer if we don't spell these methods as extensions on the group. I would prefer if we just make them private methods on `ServiceGroup` itself.

Sources/ServiceLifecycle/ServiceGroup.swift
```diff
@@ -779,6 +847,43 @@ public actor ServiceGroup: Sendable, Service {
...
+                } catch {
+                    return .serviceThrew(service: serviceConfiguration, index: index, error: error)
+                }
+            }
+        }
+    }
+
+    mutating func addAddedServiceListenerTask(_ channel: AsyncChannel<ServiceGroupConfiguration.ServiceConfiguration>) {
+        self.addTask {
+            return await withTaskCancellationHandler {
```
FranzBusch, line 872: This still seems wrong. We need to figure out why we need this. `AsyncChannel` does handle cancellation itself normally.

Sources/ServiceLifecycle/ServiceGroup.swift
```diff
@@ -717,6 +780,11 @@ public actor ServiceGroup: Sendable, Service {
                     // We are going to continue the result loop since we have to wait for our service
                     // to finish.
                     break
+
+                case .newServiceAdded:
+                    // TBD: How do we treat added services during graceful shutdown?
+                    // Currently, we ignore them - but we make sure that `run` is never called
+                    break
```
FranzBusch, line 787: I think that's fine. If we are shutting down we can just ignore them.

Sources/ServiceLifecycle/ServiceGroup.swift
```diff
@@ -310,12 +334,48 @@ public actor ServiceGroup: Sendable, Service {
...
+            }
+
+            // Adds a task that listens to added services and funnels them into the task group
+            group.addAddedServiceListenerTask(addedServiceChannel)
+
             // We are going to wait for any of the services to finish or
             // the signal sequence to throw an error.
             while !group.isEmpty {
-                let result: ChildTaskResult? = try await group.next()
+                let nextEvent = try await group.next()
```
FranzBusch, line 350: Any reason why we need to rename this?

sliemeobn: nope, none at all

Sources/ServiceLifecycle/ServiceGroup.swift
sliemeobn: done

Sources/ServiceLifecycle/ServiceGroup.swift
sliemeobn: It might be as designed though. especially with out multi-iterator `next`ing off one at a time setup, how else is the channel supposed to know that it is no longer listened to?

sliemeobn:
thanks @FranzBusch for the quick reviews!

I addressed the feedback from above, but I am still a bit scared of the "best effort" nature of it.

So, as a suggestion, I renamed the function to `addServiceUnlessShutdown` to better communicate its best-effort nature (similar to `TaskGroup.addTaskUnlessCanceled`).

This would also allow us to one day add a `addService` that maybe always runs and immediately shuts down, and maybe throws if the group was canceled or has shutdown. (or similar, should the need arise)

wdyt?

Sources/ServiceLifecycle/ServiceGroup.swift
FranzBusch: Oh thanks for this comment. Yes I remember now that async channel ignores cancellation on both sides since it is a multi producer multi consumer channel.

Sources/ServiceLifecycle/ServiceGroup.swift
```diff
@@ -777,6 +850,45 @@ public actor ServiceGroup: Sendable, Service {
...
+        group.addTask {
+            return await withTaskCancellationHandler {
+                var iterator = channel.makeAsyncIterator()
+                if let addedService = await iterator.next() {
+                    return .newServiceAdded(addedService)
+                }
+
+                return .gracefulShutdownFinished
+            } onCancel: {
+                // Without this we can get stuck in `addService` if the group
```
FranzBusch, line 887: Can we change this comment to reflect our discussion above and just say `AsyncChannel` is MPSC hence we need to shut it down on cancel.

FranzBusch (commented):
This looks great. I'm happy with the new name and the implementation looks good to me. Just one last nit around the comment

FranzBusch:
@sliemeobn can you fix the license header check please

sliemeobn:
> can you fix the license header check please
@FranzBusch oops, done

(is there an easy way to run the soundness checks locally?)

FranzBusch (approved):

FranzBusch:
> > can you fix the license header check please
> 
> @FranzBusch oops, done
> 
> (is there an easy way to run the soundness checks locally?)

Yeah you can use `act` to run all of it locally. There is some documentation here: https://github.com/swiftlang/github-workflows?tab=readme-ov-file#running-workflows-locally

Sources/ServiceLifecycle/ServiceGroup.swift
```diff
@@ -106,6 +108,38 @@ public actor ServiceGroup: Sendable, Service {
         self.maximumCancellationDuration = configuration._maximumCancellationDuration
     }
 
+    /// Adds a new service to the group.
+    ///
+    /// If the group is currently running, the added service will be started immediately.
+    /// If the group is gracefully shutting down, cancelling, or already finished, the added service will not be started.
+    /// - Parameters:
```
czechboy0, line 115: Apologies for the late review - it'd be great to call out that this can also be used for adding services before the group was started. Right now the comment leaves that ambiguous of what happens.

PR #200, opened 2025-03-10
Add preamble service that runs closure before running service

adam-fowler:
I find I am writing this small helper service across many applications, so thought it might be of use as part of the library. Basically it is a service that runs a preamble closure before running a child service. eg 

```swift
let serviceGroup = ServiceGroup(
    services: [
        PreambleService(service: webserver) { try await performDatabaseMigrations() }
    ],
    logger: logger
)
```

I have also extended it to take an array of services if the child service is a ServiceGroup

```swift
let serviceGroup = ServiceGroup(
    services: [
        PreambleService(services: [webserver, jobQueue]) { try await performDatabaseMigrations() }
    ],
    logger: logger
)
```

Struggling to find a name for the type that I'm happy with though. This was the best I could come up with.

### DIFF
--- a/Sources/ServiceLifecycle/PreambleService.swift
+++ b/Sources/ServiceLifecycle/PreambleService.swift
@@ -1,0 +1,64 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftServiceLifecycle open source project
+//
+// Copyright (c) 2025 Apple Inc. and the SwiftServiceLifecycle project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftServiceLifecycle project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+import Logging
+
+/// Service that runs a preamble closure before running a child service
+public struct PreambleService<S: Service>: Service {
+    let preamble: @Sendable () async throws -> Void
+    let service: S
+
+    ///  Initialize PreambleService
+    /// - Parameters:
+    ///   - service: Child service
+    ///   - preamble: Preamble closure to run before running the service
+    public init(service: S, preamble: @escaping @Sendable () async throws -> Void) {
+        self.service = service
+        self.preamble = preamble
+    }
+
+    public func run() async throws {
+        try await preamble()
+        try await service.run()
+    }
+}
+
+extension PreambleService where S == ServiceGroup {
+    ///  Initialize PreambleService with a child ServiceGroup
+    /// - Parameters:
+    ///   - services: Array of services to create ServiceGroup from
+    ///   - logger: Logger used by ServiceGroup
+    ///   - preamble: Preamble closure to run before starting the child services
+    public init(services: [Service], logger: Logger, _ preamble: @escaping @Sendable () async throws -> Void) {
+        self.init(
+            service: ServiceGroup(configuration: .init(services: services, logger: logger)),
+            preamble: preamble
+        )
+    }
+
+    ///  Initialize PreambleService with a child ServiceGroup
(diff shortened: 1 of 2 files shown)


czechboy0:
Hi @adam-fowler,

do you use this pattern because the contents of the preamble closure _need_ to run inside a ServiceGroup, for example because they use graceful shutdown?

I assume there's a reason this wouldn't work for you?

```swift
try await performDatabaseMigrations()
let serviceGroup = ServiceGroup(
    services: [
        webserver
    ],
    logger: logger
)
```

FranzBusch:
I think we can model this a bit differently to support `Services` either having finished or expressing some state before we start the next services. Though one thing that I'm thinking of more and more lately is that all of this can be expressed using `with-style` methods. Maybe its time we provide ways to tie in `with-style` methods into the graceful shutdown ordering.

adam-fowler:
> do you use this pattern because the contents of the preamble closure _need_ to run inside a ServiceGroup, for example because they use graceful shutdown?
> 
> I assume there's a reason this wouldn't work for you?
> 
> ```swift
> 
> ```

The database is also a service. So it needs to be running as well.

adam-fowler:
> I think we can model this a bit differently to support `Services` either having finished or expressing some state before we start the next services. Though one thing that I'm thinking of more and more lately is that all of this can be expressed using `with-style` methods. Maybe its time we provide ways to tie in `with-style` methods into the graceful shutdown ordering.

Can you give us a code example to show what you mean. `with-style` methods tend to be used for scoping resources to the confines of a closure. I can't see how that works in an array of services.

czechboy0:
> > do you use this pattern because the contents of the preamble closure _need_ to run inside a ServiceGroup, for example because they use graceful shutdown?
> > I assume there's a reason this wouldn't work for you?
> 
> The database is also a service. So it needs to be running as well.

🤦‍♂️ I completely missed that `databaseService`, thanks.

And `performDatabaseMigrations` can't run as part of `databaseService`? The order of when `databaseService` and `webserver` come up is undefined, so I assume the way `webServer` uses the database already must have a way to wait for it to start processing requests?

adam-fowler:
> 🤦‍♂️ I completely missed that `databaseService`, thanks.
> 
That's because I edited it

> And `performDatabaseMigrations` can't run as part of `databaseService`? The order of when `databaseService` and `webserver` come up is undefined, so I assume the way `webserver` uses the database already must have a way to wait for it to start processing requests?

The way I see it databaseService comes from a library eg PostgresClient from PostgresNIO and the migrations are specific to the application, so they are two separate things

czechboy0:
If the database should never be used until migrations have run, an alternative spelling would be for the `Service` implementation wrapping `PostgresClient` to offer a "postamble" closure, executed once the client is connected, but _before_ it starts processing incoming requests. But if the database _is_ meant to be used even before migrations, then this suggestion wouldn't work.

adam-fowler:
> If the database should never be used until migrations have run, an alternative spelling would be for the `Service` implementation wrapping `PostgresClient` to offer a "postamble" closure

This wouldn't work as all Services start at the same time.

FranzBusch:
In general, I think we have three approaches for the problem at hand:

- `PreambleService`: This is your PR here and we would nest ServiceGroups into each other to make sure some stuff ran before the nested service or service group starts
- Introduce a `StatefulService` protocol that services can conform to so that you can say only start the next service after a certain state of the previous has been hit
- `wit-style` graceful shutdown: In the end, both of the above are actually solved in structured concurrency however our `run` based approach leads us down to reinvent linear code flow.

I understand the pragmatism behind your proposed PR here but I'm not sure this is the overall approach that we should recommend. I would like us to think through the other design spaces as well before adding the `PreambleService` here. As you said this is something that's completely solvable in user space right now so not blocking anyone.

PR #203, opened 2025-04-23
Remove platforms from manifest

glbrntt:
Motivation:

Adding or raising the deployment platforms in the package manifest is a SemVer major breaking change as consumers must also add or raise their deployment platforms. This is a known limitation of SwiftPM.

Unforunately this means that it's very difficult for non-leaf packages to adopt packages which declare their platforms in the manifest. Doing so puts the brakes on adoption and ecosystem growth. For 'core' packages like this one availability constraints should be expressed on declarations rather than in the manifest.

Modifications:

- Remove platforms from the package manifest
- Add availability annotations to types which require it

Result:

This package can be more widely adopted

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -3,12 +3,6 @@ import PackageDescription
 
 let package = Package(
     name: "swift-service-lifecycle",
-    platforms: [
-        .macOS(.v10_15),
-        .iOS(.v13),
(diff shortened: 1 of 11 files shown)


FranzBusch (approved):
Amazing. Thanks!

PR #205, opened 2025-05-05
adding abstracts and organization to methods

heckj:
A little bit of docs cleanup - adding abstracts for deprecated methods and curating them (organizing them) to the bottom of the relevant lists.


heckj:
I'm not 100% sure this will pass soundness checks, as I think some of what I added in terms of disambiguating overloads uses a Swift 6.1 method, and may not work smoothly with Swift 6.0 - if we could trigger the CI, that would let me know pretty quickly. I don't anticipate any other issues though.

FranzBusch (approved):
Great stuff! Can you fix up the formatting?

PR #215, opened 2025-09-22
Enable 6.2 CI and enable Windows CI

FranzBusch:


.github/workflows/pull_request.yml
```diff
@@ -15,12 +15,12 @@ jobs:
...
     with:
-      linux_5_9_arguments_override: "--explicit-target-dependency-import-check error"
-      linux_5_10_arguments_override: "--explicit-target-dependency-import-check error"
       linux_6_0_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error -Xswiftc -require-explicit-sendable"
       linux_6_1_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error -Xswiftc -require-explicit-sendable"
-      linux_nightly_6_0_arguments_override: "--explicit-target-dependency-import-check error -Xswiftc -require-explicit-sendable"
+      linux_6_2_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error -Xswiftc -require-explicit-sendable"
+      linux_nightly_next_arguments_override: "--explicit-target-dependency-import-check error -Xswiftc -require-explicit-sendable"
       linux_nightly_main_arguments_override: "--explicit-target-dependency-import-check error -Xswiftc -require-explicit-sendable"
+      windows_6_0_enabled: true
```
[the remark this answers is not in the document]
czechboy0: Windows is disabled by default.

czechboy0 (approved):

Sources/UnixSignals/UnixSignal.swift
```diff
@@ -26,6 +26,8 @@ import Dispatch
 /// A struct representing a Unix signal.
 ///
 /// Signals are standardized messages sent to a running program to trigger specific behavior, such as quitting or error handling
+///
+/// - Important: Unix signals are only functional on platforms supporting signals.
```
czechboy0, line 30: Should we instead compile out these unix signals on Windows for now completely, so even the ServiceGroup doesn't have those parameters in its initializers? We can later add support for the Windows-specific termination handling using the right terminology, and add more initializers then.

czechboy0: Once we make this build on Windows, "UnixSignal" becomes API on Windows. I don't love that and would prefer we compile it out and only deliberately add a potential Windows-specific or a cross-platform API (called e.g. "service event") that uses the right platform-specific API under the hood. But landing this is a one-way door we can't remove for the rest of ServiceLifecycle 2.x, and we don't really benefit.

I think it's fine for those writing cross-platform services to handle this in their app for now, and we can introduce a cross-platform event API later.

czechboy0: Would `WindowsConsoleEvent` be a symbol that's present on macOS and Linux?

simonjbeaumont: Have you considered moving unix signals behind a trait. Trait would be default enabled on all platforms where Swift Service Lifecycle currently builds. Could default to disabled on Windows? Could possibly even annotate them as unavailable?

czechboy0: > Yes. Same argument applies the other way around. Somebody wanting to write a cross platform service that uses ServiceGroup should not be forced to write #if os() code.

Fair enough, while I don't agree with the tradeoff here, it is a reasonable position and I'll defer to you here.

FranzBusch:
@Joannis LGTM! Thanks for fixing this. Merging the PR

PR #217, opened 2025-10-07
Include the service name in the `serviceFinishedUnexpectedly` error

FranzBusch:
## Motivation

To aid debugging and troubleshooting, it is helpful to understand which Service led to the error.

## Modifications

Add the service name to the `serviceFinishedUnexpectedly` error.

## Result

Easier to debug failing services.

### DIFF
--- a/Sources/ServiceLifecycle/ServiceRunnerError.swift
+++ b/Sources/ServiceLifecycle/ServiceRunnerError.swift
@@ -50,14 +50,16 @@ public struct ServiceGroupError: Error, Hashable, Sendable {
 
     /// Internal class that contains the actual error code.
     private final class Backing: Hashable, Sendable {
+        let message: String?
         let errorCode: Code
         let file: String
         let line: Int
 
-        init(errorCode: Code, file: String, line: Int) {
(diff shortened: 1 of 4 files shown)


Sources/ServiceLifecycle/ServiceRunnerError.swift
```diff
@@ -50,14 +50,16 @@ public struct ServiceGroupError: Error, Hashable, Sendable {
 
     /// Internal class that contains the actual error code.
     private final class Backing: Hashable, Sendable {
+        let message: String?
         let errorCode: Code
         let file: String
         let line: Int
 
-        init(errorCode: Code, file: String, line: Int) {
+        init(errorCode: Code, file: String, line: Int, message: String? = nil) {
```
czechboy0, line 58:
```suggestion
        init(errorCode: Code, file: String, line: Int, message: String?) {
```

Nit: I'd drop the default

czechboy0 (approved):
One suggestion, otherwise lgtm

FranzBusch:
Merging over the API break since this is actually an ABI break

PR #252, opened 2026-08-04
Default ServiceGroup logger to Logger.current

### LINKED ISSUE #251: Default the logger parameter to Logger.current
Now that we have structured logging, don't force every adopter to explicitly pass a logger in the initializer of the ServiceGroup - default to `.current` instead.

ZHUOLIN0928:
Fixes #251

## Summary
- Raise the minimum swift-log dependency to 1.14.0, which provides the task-local Logger.current API.
- Default ServiceGroup's logger parameter to Logger.current so callers can inherit the current task-local logger.
- Add a regression test that constructs and runs ServiceGroup without an explicit logger.

## Design
The change is limited to the current ServiceGroup convenience initializer. Existing explicit logger calls and ServiceGroupConfiguration initializers remain unchanged.

## Validation
- Focused test passed: swift test --disable-sandbox --filter ServiceGroupTests/testRun_withDefaultLogger
- Full test suite passed: swift test --disable-sandbox
- Result: 62 ServiceLifecycle/XCTest tests and all Swift Testing suites passed.
- git diff --check passed.

Tests used the Xcode 27 beta developer toolchain because the active Command Line Tools installation does not provide XCTest.


FranzBusch (approved):

Tests/ServiceLifecycleTests/ServiceGroupTests.swift
```diff
@@ -60,6 +60,12 @@ final class ServiceGroupTests: XCTestCase {
         try await group.run()
     }
 
+    func testRun_withDefaultLogger() async throws {
+        let group = ServiceGroup(services: [])
+
+        try await group.run()
```
czechboy0, line 66: This should probably be wrapped in withLogger and use the in-memory logger to verify the expected Logger was used. Not sure what this test is testing otherwise.

FranzBusch: I think a second test for that would be good. When I reviewed this I read this test as "You can now create a group without passing a logger" which wasn't possible before.

czechboy0: @ZHUOLIN0928 Is this something you'd be able to pick up?

PR #258, opened 2026-09-11
Fix dispatch source leaks in UnixSignalsSequence

### LINKED ISSUE #256: Signal source remains allocated after dropping UnixSignalsSequence on macOS
While testing SIGHUP support for apple/swift-configuration#20, I noticed that a dispatch signal source remains allocated after a `UnixSignalsSequence` and its iterator go out of scope. Repeating this leaves one additional live source per iteration on macOS.

I saw the documentation's warning about multiple handlers for the same signal. My use case is one sequence per running service, created inside `Service.run()`. The reproducer drops each sequence and iterator before creating the next. I would expect the source to be released when that ownership ends, including when a sequence is only created once.

### Environment

- SwiftServiceLifecycle 2.12.0, commit `7f9326b0326ff86e3646295ea6e891f68c471c5e`
- macOS 26.6.2 (25G83), arm64
- Apple Swift 6.3.3 (`swiftlang-6.3.3.1.3`, target `arm64-apple-macosx26.0`)

### Reproduction

Create a directory containing these two files. Each iteration creates a sequence, sends SIGHUP to the process, receives one signal, and lets the sequence and iterator go out of scope. The program then uses `/usr/bin/heap` to count the remaining dispatch sources.

`Package.swift`:

```swift
// swift-tools-version: 6.0
import PackageDescription

let package = Package(
    name: "DispatchSourceLeakProbe",
    platforms: [.macOS(.v13)],
    dependencies: [
        .package(url: "https://github.com/swift-server/swift-service-lifecycle.git", exact: "2.12.0")
    ],
    targets: [
        .executableTarget(
            name: "DispatchSourceLeakProbe",
            dependencies: [
                .product(name: "UnixSignals", package: "swift-service-lifecycle")
            ]
        )
    ]
)
```

`Sources/DispatchSourceLeakProbe/main.swift`:

```swift
import Darwin
import Foundation
import UnixSignals

let count = CommandLine.arguments.dropFirst().first.flatMap(Int.init) ?? 1_000

for _ in 0..<count {
    let signals = await UnixSignalsSequence(trapping: .sighup)
    precondition(kill(getpid(), SIGHUP) == 0)
    for await _ in signals {
        break
    }
}

// Give the dispatch queue a moment to settle before inspecting the heap.
try await Task.sleep(for: .milliseconds(200))

let heap = Process()
heap.executableURL = URL(fileURLWithPath: "/usr/bin/heap")
heap.arguments = ["\(getpid())"]
let pipe = Pipe()
heap.standardOutput = pipe
try heap.run()
let output = String(decoding: pipe.fileHandleForReading.readDataToEndOfFile(), as: UTF8.self)
heap.waitUntilExit()

let sourceLines = output.split(separator: "\n").filter { $0.contains("dispatch_source") }
print("Created and dropped \(count) UnixSignalsSequence instances.")
print("Live dispatch sources according to heap:")
for line in sourceLines {
    print("  \(line.trimmingCharacters(in: .whitespaces))")
}
```

Run the control and reproduction from that directory:

```sh
swift run -c release DispatchSourceLeakProbe 0
swift run -c release DispatchSourceLeakProbe 2000
```

### Expected and actual results

I would expect the sources to be cancelled and released once their sequences and iterators are no longer retained, rather than accumulating across iterations.

The control reports 5 live `dispatch_source_t` objects. After 2,000 iterations, it reports 2,005:

```text
Created and dropped 2000 UnixSignalsSequence instances.
Live dispatch sources according to heap:
  2005     256640     128.0   dispatch_source_t   ObjC   libdispatch.dylib
```

I reproduced this in a release build on the environment above. I have not measured it on Linux.

In [`UnixSignalsSequence.Storage`](https://github.com/swift-server/swift-service-lifecycle/blob/7f9326b0326ff86e3646295ea6e891f68c471c5e/Sources/UnixSignals/UnixSignalsSequence.swift#L111), the event handler captures `source`, which holds the dispatch source. The calls to `cancel()` appear to cover cancellation during initialization, but not normal teardown. This looks like a retain-cycle or cleanup issue.

I'm happy to work on a fix.

mmarufov:
Fixes #256.

## Motivation

`UnixSignalsSequence` keeps dispatch signal sources alive after the sequence and its iterator are released. The event handler captures the whole `Source` value, which holds the `DispatchSource`, creating a retain cycle.

## Modifications

- Keep the sources on `Storage` and cancel them in `deinit`.
- Resume sources skipped when initialization is cancelled so libdispatch can run cancellation cleanup and release their handlers.
- Add regression tests for normal teardown and cancelled initialization. The teardown test waits for cancellation because the registration callback can briefly retain `Storage` after initialization returns.

## Result

Sources are cleaned up after the last sequence or iterator releases the storage, including when initialization is cancelled.

In the macOS reproducer, 2,000 create/drop iterations went from 2,005 live dispatch sources before the fix to 5 afterward, matching the zero-iteration control. These measurements were made on macOS only.

All 10 UnixSignals tests pass locally in release mode on macOS with Swift 6.3.3 and the CI warning, explicit-import, and Sendable flags. Strict formatting and `git diff --check` also pass. The weak-reference test has not been run locally on Linux; its first Linux validation will be in CI.


Sources/UnixSignals/UnixSignalsSequence.swift
```diff
@@ -82,10 +82,32 @@ public struct UnixSignalsSequence: AsyncSequence, Sendable {
...
+        fileprivate let sources: [Source]
+
+        deinit {
+            // Dispatch sources retain their event handlers until they are cancelled, and the
+            // handler holds on to the source, so we have to cancel them here to avoid leaking
+            // a registered signal source per instance.
+            for source in self.sources {
+                source.dispatchSource.cancel()
+            }
+        }
```
FranzBusch, line 109: This is obv the right fix for now since this already shipped but can you please file a follow up issue that we should respell this API to be with-style and deprecate the init one:

```swift
await withUnixSignals([.sigterm]) { signal
 // handle signal
}
```

This way we can follow structured resource management

FranzBusch (approved):

Sources/UnixSignals/UnixSignalsSequence.swift
mmarufov: Opened #259 to track the with-style API follow-up. Thanks for the review!

